PR #3, opened 2020-05-27
Build and Test using GitHub Actions

jgiannuzzi:
* Add a CI pipeline to build and test using GitHub Actions
* Fix building and testing on Unix, where .NET Framework is not available
* Repack on .NET Framework to avoid Newtonsoft.Json dependency
* Generate Authors and Description for the NuGet package
* Add a version suffix with the short commit when building for a branch


jgiannuzzi:
Here is an example run: https://github.com/jgiannuzzi/consuldotnet/actions/runs/117063291

.github/workflows/ci.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          cd Consul.Test
+          SYSTEM=$(uname | sed 's/MINGW.*/windows/' | tr A-Z a-z)
+          curl -sSL https://releases.hashicorp.com/consul/${CONSUL_VERSION}/consul_${CONSUL_VERSION}_${SYSTEM}_amd64.zip -o consul.zip
+          unzip consul.zip
+          rm consul.zip
+      - name: Build
+        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true
+      - name: Repack
+        if: runner.os == 'Windows'
```
GPSnoopy, line 35: Could use a comment explaining why you're doing this.

jgiannuzzi: Good point, I added comments to the CI build job.
FYI this is preserving existing behaviour of the previous CI.

PR #4, opened 2020-05-29
Tidy up lock tests and increase timeouts

adamreeve:
* Increases timeouts in lock tests
* Fixes exception checks to actually check an exception is thrown
* Fix tests that didn't actually test anything
* Various tidy ups to make tests more readable

I still need to look at the semaphore tests and make similar changes.

### DIFF
--- a/Consul.Test/LockTest.cs
+++ b/Consul.Test/LockTest.cs
@@ -300,7 +300,10 @@ namespace Consul.Test
                 }));
             }
 
-            await WithTimeout(Task.WhenAll(tasks));
+            // Set timeout to allow each task to wait up to the lock wait time
+            await WithTimeout(
+                Task.WhenAll(tasks),
+                TimeSpan.FromTicks(2 * contenderPool * Lock.DefaultLockWaitTime.Ticks));
 
             for (var i = 0; i < contenderPool; i++)
             {
--- a/Consul.Test/LockTest.cs
+++ b/Consul.Test/LockTest.cs
@@ -134,9 +134,10 @@ namespace Consul.Test
 
             Assert.True(lockKey.IsHeld);
             Assert.False(contender.IsHeld);
-            Assert.NotEqual(WaitHandle.WaitTimeout, Task.WaitAny(
-                new Task[] { Task.Run(async () => { await Assert.ThrowsAsync<LockMaxAttemptsReachedException>(async () => await contender.Acquire()); }) },
-                (int)(2 * lockOptions.LockWaitTime.TotalMilliseconds)));
+
+            await WithTimeout(
+                Assert.ThrowsAsync<LockMaxAttemptsReachedException>(async () => await contender.Acquire()),
+                TimeSpan.FromMinutes(1));
 
             Assert.False(stopwatch.ElapsedMilliseconds < lockOptions.LockWaitTime.TotalMilliseconds);
             Assert.False(contender.IsHeld, "Contender should have failed to acquire");
@@ -177,9 +178,8 @@ namespace Consul.Test
             Assert.True(l.IsHeld);
             await _client.Session.Destroy(sessionId.Response);
 
-            await Task.Delay(TimeSpan.FromSeconds(1));
+            await WaitFor(() => !l.IsHeld, "Expected lock to be lost when session destroyed", TimeSpan.FromMinutes(1));
 
-            Assert.False(l.IsHeld);
             Assert.Null((await _client.KV.Get(keyName)).Response);
         }
 
@@ -269,9 +269,7 @@ namespace Consul.Test
                 }));
             }
 
-            await Task.WhenAny(
-                Task.WhenAll(tasks),
-                Task.Delay(TimeSpan.FromTicks(contenderPool * Lock.DefaultLockWaitTime.Ticks)));
+            await WithTimeout(Task.WhenAll(tasks), TimeSpan.FromMinutes(5));
 
             for (var i = 0; i < contenderPool; i++)
             {
@@ -303,9 +301,7 @@ namespace Consul.Test
                 }));
             }
 
-            await Task.WhenAny(
-                Task.WhenAll(tasks),
-                Task.Delay(TimeSpan.FromTicks(contenderPool * Lock.DefaultLockWaitTime.Ticks)));
+            await WithTimeout(Task.WhenAll(tasks), TimeSpan.FromMinutes(1));
 
             for (var i = 0; i < contenderPool; i++)
             {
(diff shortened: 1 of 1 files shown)


Consul.Test/LockTest.cs
```diff
@@ -177,9 +178,8 @@ public async Task Lock_EphemeralAcquireRelease()
             Assert.True(l.IsHeld);
             await _client.Session.Destroy(sessionId.Response);
 
-            await Task.Delay(TimeSpan.FromSeconds(1));
+            await WaitFor(() => !l.IsHeld, "Expected lock to be lost when session destroyed", TimeSpan.FromMinutes(1));
```
jgiannuzzi, line 181: What is the rationale for waiting up to 1 minute in this case?

Consul.Test/LockTest.cs
```diff
@@ -269,9 +269,7 @@ public async Task Lock_ContendWait()
                 }));
             }
 
-            await Task.WhenAny(
-                Task.WhenAll(tasks),
-                Task.Delay(TimeSpan.FromTicks(contenderPool * Lock.DefaultLockWaitTime.Ticks)));
+            await WithTimeout(Task.WhenAll(tasks), TimeSpan.FromMinutes(5));
```
jgiannuzzi, line 272: What is the rationale for waiting up to 5 minutes in this case, instead of 1 minute like in most other tests?

adamreeve:
Thanks Jonathan, I've added a default timeout value and added a comment on this and a comment where the longer timeout is used. I switched `Lock_ContendWait` to the default instead of 5 minutes as this should only take 3 seconds usually.

adamreeve:
Hmm I had another failure on my fork in Lock_ContendFast, looks like it was wrong to only have a minute timeout there. It was originally set to allow for a wait of DefaultLockWaitTime per task (150 s) although was always completing within a second for me locally. I'll bump that back up again.

Consul.Test/LockTest.cs
```diff
@@ -300,7 +300,10 @@ public async Task Lock_ContendFast()
                 }));
             }
 
-            await WithTimeout(Task.WhenAll(tasks));
+            // Set timeout to allow each task to wait up to the lock wait time
+            await WithTimeout(
+                Task.WhenAll(tasks),
+                TimeSpan.FromTicks(2 * contenderPool * Lock.DefaultLockWaitTime.Ticks));
```
jgiannuzzi, line 306: Now that we are using absolute timeouts everywhere else, why do we use a computed one here?
I think we should be consistent and either compute them everywhere, or nowhere. What do you think?

jgiannuzzi, line 306: Now that we are using absolute timeouts everywhere else, why do we use a computed one here?
I think we should be consistent and either compute them everywhere, or nowhere. What do you think?

adamreeve: Yeah good point, I think in the cases where the default timeout is overridden it probably makes sense to compute it but with some allowance for extra time rather than just setting it to the maximum expected time as it was before, so I should probably also compute the timeout for `Lock_Contend_LockDelay`.

Although because the timeout is really just there to catch a hang rather than as any kind of performance check, I think it might just make sense to bump up the default timeout to something larger like 5 or 10 minutes and use that everywhere. Would that seem reasonable to you?

adamreeve:
Hmm got another failure I hadn't seen before:
```
2020-06-01T08:29:52.6407219Z   X Consul.Test.SemaphoreTest.Semaphore_OneShot [6s 835ms]
2020-06-01T08:29:52.6407618Z   Error Message:
2020-06-01T08:29:52.6407991Z    Consul.SemaphoreHeldException : Exception of type 'Consul.SemaphoreHeldException' was thrown.
2020-06-01T08:29:52.6408334Z   Stack Trace:
2020-06-01T08:29:52.6408709Z      at Consul.Semaphore.<Destroy>d__29.MoveNext() in D:\a\consuldotnet\consuldotnet\Consul\Semaphore.cs:line 539
2020-06-01T08:29:52.6409297Z --- End of stack trace from previous location where exception was thrown ---
2020-06-01T08:29:52.6409687Z    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
2020-06-01T08:29:52.6410400Z    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
2020-06-01T08:29:52.6410777Z    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
2020-06-01T08:29:52.6411143Z    at Consul.Test.SemaphoreTest.<Semaphore_OneShot>d__6.MoveNext() in D:\a\consuldotnet\consuldotnet\Consul.Test\SemaphoreTest.cs:line 186
2020-06-01T08:29:52.6411588Z --- End of stack trace from previous location where exception was thrown ---
2020-06-01T08:29:52.6411957Z    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
2020-06-01T08:29:52.6412331Z    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
2020-06-01T08:29:52.6412766Z --- End of stack trace from previous location where exception was thrown ---
2020-06-01T08:29:52.6413251Z    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
2020-06-01T08:29:52.6413661Z    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
2020-06-01T08:29:52.6414083Z --- End of stack trace from previous location where exception was thrown ---
2020-06-01T08:29:52.6414328Z    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
2020-06-01T08:29:52.6414566Z    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```
Looks like a race condition when destroying the semaphore.

Consul.Test/LockTest.cs
jgiannuzzi: > Although because the timeout is really just there to catch a hang rather than as any kind of performance check, I think it might just make sense to bump up the default timeout to something larger like 5 or 10 minutes and use that everywhere. Would that seem reasonable to you?

That does sound reasonable to me. Let's go for a default of 5 minutes and add a comment that this is only there to catch a hang, and not a performance check. If we do not need those utility methods to ever be called with a timeout different from the default, then let's remove the optional parameter.

PR #5, opened 2020-06-01
Tidy up semaphore tests

adamreeve:
* Move timeout utils into a separate class
* Fix tests that check for exceptions being thrown
* Use longer timeouts when waiting for tasks to reduce flakiness in CI
* Fix Semaphore_ExecuteAction so it actually checks the action is executed
* Remove Semaphore_ForceInvalidate test as this wasn't actually testing anything different to Semaphore_Destroy

Contains the same changes as #4 so may need rebasing once that is merged


jgiannuzzi:
Rebased after merging #4

PR #14, opened 2020-06-08
Add NuGet symbol package

jgiannuzzi:
See https://devblogs.microsoft.com/nuget/improved-package-debugging-experience-with-the-nuget-org-symbol-server/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
           } else {
             $params = "--version-suffix", $(git rev-parse --short HEAD)
           }
-          dotnet pack --configuration=Release --no-build @params
+          dotnet pack --configuration=Release --no-build --include-symbols -p:SymbolPackageFormat=snupkg @params
       - name: Upload NuGet package artifact
         if: runner.os == 'Windows'
         uses: actions/upload-artifact@v2


mfkl:
Personally I'm more fan of shipping the pdb inside the nuget than using a symbol server. Less reliance on internet and nuget infra. 

Just my 2 cents :-)

skyguy94:
Its probably worth considering official guidance on when to include a pdb in a package:

> Including PDBs in the .nupkg is generally no longer recommended as it increases the size of the package and thus restore time for projects that consume your package, regardless of whether the user needs to debug through the source code of your library or not. That said, .snupkg symbol packages have some limitations:
> * They do not currently support Windows PDBs (generated by VC++, or for managed projects that set build property DebugType to full)
>  * They require the library to be built by newer C#/VB compiler (Visual Studio 2017 Update 9).
>  * The consumer of the package also needs Visual Studio 2017 Update 9 debugger.
>  * Not supported by Azure DevOps Artifacts service.
>Consider including PDBs in the main package only if it is not possible to use .snupkg for the above reasons. For managed projects, consider switching to Portable PDBs by setting DebugType property to portable. This is the default for .NET SDK projects, but not classic .NET projects.

Link: [Alternative PDB Distribution](https://github.com/dotnet/sourcelink#alternative-pdb-distribution)

That being said Sourcelink is a nice feature for package source debugging:  https://github.com/dotnet/sourcelink

marcin-krystianc:
I would prefer shipping pdbs with the nuget package itself instead of using `snupkg` format as the latter is not yet supported in Rider (https://youtrack.jetbrains.com/issue/RIDER-21735) and in Artifactory (https://www.jfrog.com/jira/browse/RTFACT-5972).

sungam3r:
> as it increases the size of the package 

Not so much since portable embedded pdbs.

> I would prefer shipping pdbs with the nuget package itself instead of using snupkg format

So do I. In fact package size increases slightly: 579K now without embedded PDBs and 626K with embedding PDBs.

PR #17, opened 2020-06-17
Add initial ASP.NET extensions project

sergeyshaykhullin:
Add `Consul.AspNetCore` project. Named like another `AspNetCore` integration libraries:
- https://www.nuget.org/packages/Swashbuckle.AspNetCore
- https://www.nuget.org/packages/Serilog.AspNetCore
- https://www.nuget.org/packages/Hangfire.AspNetCore
- https://www.nuget.org/packages/FluentValidation.AspNetCore
- https://www.nuget.org/packages/HotChocolate.AspNetCore

I need help with:
- Supported targets (netstandard2.0?)
- Nuget publish (CI?)
- Package descriptions
- Service register/deregister design


jgiannuzzi:
Thanks for creating this PR, @sergeyshaykhullin! 🚀

`Consul.AspNetCore` sounds good to me.

> I need help with:
> 
> * Supported targets (netstandard2.0?)

I think `netstandard2.0` is the right target, yes. The tests should target both .NET Core 2.0 and 3.1, like the main project.

> * Nuget publish (CI?)

We can take care of the CI in a separate PR, once this one is done.

> * Package descriptions

I am thinking of "Consul Service registration for ASP.NET Core".

> * Service register/deregister design

I don't have enough experience with ASP.NET Core to comment on that one. Anyone?

jgiannuzzi:
Could you also please add a test project `Consul.AspNetCore.Test` with the relevant tests?

sergeyshaykhullin:
@jgiannuzzi I am not used this constructor override, but it is useful using DI containers. So add configuration to DI container, inject it to ConsulClient
```cs
[Obsolete("This constructor is no longer necessary due to the new Action based constructors and will be removed in a future release." +
            "Please use the ConsulClient(Action<ConsulClientConfiguration>) constructor to set configuration options.", false)]
        public ConsulClient(ConsulClientConfiguration config)
```
Then i can resolve configuration from DI and assert in tests instead of casting and getting properties.

P.S. I used implementation factory insted, it works fine

Consul.AspNetCore.Test/Consul.AspNetCore.Test.csproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">netcoreapp2.0;netcoreapp3.1;net461</TargetFrameworks>
```
lorinck, line 4: As far as I understand this is a .Net Core project, in which case I don't think the 'net461' target is needed?

Consul.AspNetCore.Test/ServiceCollectionExtensionsTest.cs
```diff
@@ -0,0 +1,79 @@
+using System;
+using Xunit;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Consul.AspNetCore.Test
+{
+	public class ServiceCollectionExtensionsTest
+	{
+		private readonly IServiceCollection services;
```
lorinck, line 10: The naming convention is '_lowerCamelCase' for private fields in the library. Can you please start the variable name with underscore?

Consul.AspNetCore/AgentServiceRegistrationHostedService.cs
```diff
@@ -0,0 +1,30 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace Consul.AspNetCore
+{
+	public class AgentServiceRegistrationHostedService : IHostedService
+	{
+		private readonly IConsulClient consulClient;
+		private readonly AgentServiceRegistration registration;
```
lorinck, line 10: The naming convention is '_lowerCamelCase' for private fields in the library. Can you please start the variable names with underscore?

Consul.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+		public static IServiceCollection AddConsul(
+			this IServiceCollection services,
+			Action<ConsulClientConfiguration> options)
+		{
+			services.TryAddSingleton<IConsulClient>(sp => new ConsulClient(options));
+
+			return services;
+		}
+
+		public static IServiceCollection AddConsulRegistration(
```
lorinck, line 23: I find the method name a bit confusing. IMO 'AddConsulServiceRegistration' would be more descriptive and would allow extending the class later with check registration.

lorinck:
Hello @sergeyshaykhullin 

Do you mean Consul or .Net core by the service?
> * Service register/deregister design

Thanks,
Lorinc

sergeyshaykhullin:
@lorinck I mean that AspNetCore has different approaches to design "OnStart"/"OnStop" events.

1. `services.AddConsulServiceRegistration()` + `IHostedService`
2. `app.UseConsulServiceRegistration()` + `IApplicationLifetime`
3. `Host.CreateDefaultBuilder(args).UseConsulServiceRegistration()`

> * Service register/deregister design

I chose 1 options

maurosbarigia:
I tend to agree with @sergeyshaykhullin on the Obsolete constructor. We inherited it already obsoleted, but we may want to reconsider the idea. I understand Microsoft's trend of passing actions to the extension methods of IServiceCollection, but this is actually a constructor of a model class. Also, this forces ConsulClientConfiguration to be mutable, for not real benefit.
We may want to keep it obsolete if we have a good reason for preferring the constructor that takes an action, so this will flag a warning in case of explicit use in code, but leaving it available for the DI container. In this case we should change the obsolete message.

sergeyshaykhullin:
@jgiannuzzi

cecilphillip:
The folks at Steeltoe depend on this lib and also have some [aspnetcore integrations](https://github.com/SteeltoeOSS/Steeltoe/tree/3.0.0-m3/src/Discovery/src/ConsulBase). This might be an opportunity for collaboration to reduce duplication.

cc @TimHess

maurosbarigia (approved):

jgiannuzzi:
The Windows CI build is failing, but I am trying to solve it in #28

jgiannuzzi:
Thanks a lot for your contribution, @sergeyshaykhullin!

Issue #20, opened 2020-07-09
Unnecessary delay in LockTryOnce scenario


maxoris:
Currently if we are using single shot locking strategy and creating `Lock` with `LockTryOnce = true` option, in case of negative locking result we have to wait for `LockRetryTime`, before the checking for `LockTryOnce` and loop brake.
I suppose it would be great to check for value of `LockTryOnce` before introducing a delay.

jgiannuzzi:
Thanks for your interest in Consul.NET @maxoris, and sorry for the very long delay in replying! (You could call it an unnecessary delay as well 😅)

I would like to clarify the issue you are facing.
If the lock is already held when trying to acquire it, then setting `LockTryOnce = true` and `LockWaitTime = TimeSpan.FromTicks(0)` will fail immediately with `LockMaxAttemptsReachedException`. If instead a lock delay is in effect (see https://www.consul.io/docs/internals/sessions.html), and whether `LockWaitTime` is zero or non-zero, then indeed we will wait for `LockRetryTime` and then fail (unless the lock was released in the meantime).

A fix would need to check whether `LockTryOnce` is `false` and whether the elapsed time is greater than `LockWaitTime` before waiting for `LockRetryTime`. If we don't do that, I think we might introduce a hot loop.

I will review your PR #21, but I would like to ask @highlyunavailable for her valuable opinion before going further.

highlyunavailable:
This code was directly ported from the Consul Go code, so many idiosyncrasies of the Go code were inherited. You can see the [current behavior](https://github.com/hashicorp/consul/blob/master/api/lock.go#L226-L248) in said code too matches this.

If the condition was changed to `if (!Opts.LockTryOnce)` then I think it _might_ work, but as it stands now, the code in #21 actually makes a hot loop happen if LockTryOnce isn't set, as noted. **That said** I actually think this is not a correct change and `LockWaitTime = TimeSpan.Zero` (or `FromTicks(0)`) should be used as an option instead.

The reason for this is that the consul API uses HTTP long polls. These can be shortcircuited with a "WaitTime" argument on the client, which causes the Consul Agent side to drop the long poll early (after WaitTime). This causes an early return on the API.

[...]

PR #21, opened 2020-07-09
check for LockTryOnce before delay

maxoris:
Solving issue https://github.com/G-Research/consuldotnet/issues/20

### DIFF
--- a/Consul/Lock.cs
+++ b/Consul/Lock.cs
@@ -335,10 +335,13 @@ namespace Consul
                             continue;
                         }
 
-                        // If the session is null and the lock failed to acquire, then it means
-                        // a lock-delay is in effect and a timed wait must be used to avoid a hot loop.
-                        try { await Task.Delay(Opts.LockRetryTime, ct).ConfigureAwait(false); }
-                        catch (TaskCanceledException) { /* Ignore TaskTaskCanceledException */}
+                        if (Opts.LockTryOnce)
+                        {
+                            // If the session is null and the lock failed to acquire, then it means
+                            // a lock-delay is in effect and a timed wait must be used to avoid a hot loop.
+                            try { await Task.Delay(Opts.LockRetryTime, ct).ConfigureAwait(false); }
+                            catch (TaskCanceledException) { /* Ignore TaskTaskCanceledException */}
+                        }
                     }
                     DisposeCancellationTokenSource();
                     throw new LockNotHeldException("Unable to acquire the lock with Consul");


mfkl:
Should the unit tests be adjusted/modified to reflect this change? https://github.com/G-Research/consuldotnet/blob/master/Consul.Test/LockTest.cs

jgiannuzzi:
> Should the unit tests be adjusted/modified to reflect this change? https://github.com/G-Research/consuldotnet/blob/master/Consul.Test/LockTest.cs

Thanks @mfkl, that is a very good point!

I do wonder how to properly test this however, knowing that we can't rely on precise timings on the CI system.
Any ideas?

mfkl:
> I do wonder how to properly test this however, knowing that we can't rely on precise timings on the CI system.
Any ideas?

Maybe with some mocks? But it would require some more work. I'd say if the values used in tests make it pass 100% of the time, it should be fine (though not ideal). This PR needs a rebase before it is merged.

LGTM.

mfkl:
```
[xUnit.net 00:03:59.70]     Consul.Test.LockTest.Lock_TryAcquireOnceWithLockDelayNonZeroWaitTime_EnsureRetryWait [FAIL]
Error: Consul.Test.LockTest.Lock_TryAcquireOnceWithLockDelayNonZeroWaitTime_EnsureRetryWait: Assert.True() Failure
  Failed Consul.Test.LockTest.Lock_TryAcquireOnceWithLockDelayNonZeroWaitTime_EnsureRetryWait [5 s]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at Consul.Test.LockTest.Lock_TryAcquireOnceWithLockDelayNonZeroWaitTime_EnsureRetryWait() in /home/runner/work/consuldotnet/consuldotnet/Consul.Test/LockTest.cs:line 604
```

mfkl:
That [simple attempt](https://github.com/G-Research/consuldotnet/pull/21/commits/28d39642854ad9d8f4b2f6cd406ab15751bd6590) worked on my fork's CI :/

I can't repro this locally. Inclined to think the test is a bit too strict, timing-wise...

mfkl:
On WSL Ubuntu 20.04:
```
Test Run Successful.
Total tests: 172
     Passed: 168
    Skipped: 4
```
Yet it looks like I can reproduce the test failure locally on WSL Ubuntu 18.4... only _sometimes_ 🤔

mfkl:
With some trouble, I can reproduce _sporadically_ the test failure on ubuntu by letting this script run the test in a loop for a while

```
#!/bin/bash

err=0
while true
do
    dotnet test Consul.Test --configuration=Release --framework=net6.0 --logger:"console;verbosity=detailed" --filter "Lock_TryAcquireOnceWithLockDelayNonZeroWaitTime_EnsureRetryWait"
    (( (err=$?) > 0 )) && break
done
echo "$err" 
```

After several minutes, it indeed fails the test. I added debug log statements in a few places, and it turns out that this line https://github.com/G-Research/consuldotnet/blob/aa46828ead40a341461cfaaf0d391f419992cfd9/Consul/Lock.cs#L341

doesn't actually do what it's supposed to. With
```csharp
Console.WriteLine(">>>> Waiting for " + Opts.LockRetryTime.TotalMilliseconds);
Console.WriteLine(">>>> current watch before wait " + sw.ElapsedMilliseconds);
try 
{
    await Task.Delay(Opts.LockRetryTime, ct).ConfigureAwait(false); 
    Console.WriteLine(">>>> current watch after wait " + sw.ElapsedMilliseconds);
}
```

When the test fails, this actually outputs

```
>>>> Waiting for 5000
>>>> current watch before wait 2
>>>> current watch after wait 4994
```

I believe we might be hitting https://github.com/dotnet/runtime/issues/45585. Using a `Thread.Sleep` here does not sound good, even only in Linux build. We could make the test less strict about the timing through (add 10-15 milliseconds just in case), though that feels hackish as well.

marcin-krystianc:
I could also reproduce this problem locally and it seems clear that https://github.com/dotnet/runtime/issues/45585 is real (or maybe it is just an issue with documentation ?!).
My conclusions:
- On Linux, measuring elapsed time with `Environment.TickCount64`  seems to work reliably, whereas `Stopwatch` can sometimes measure smaller elapsed time than the delay itself.
- On Windows, there is no reliable technique to measure elapsed time as both `Environment.TickCount64`  and `Stopwatch` can sometimes measure smaller elapsed time than the delay itself.

Given that on Windows all existing methods of measuring elapsed time can give value smaller than the delay, I would argue that we should use some fraction of expected delay time in tests e.g.:
```
// https://github.com/dotnet/runtime/issues/45585
Assert.True(stopwatch.ElapsedMilliseconds > oneShotLockOptions.LockRetryTime.TotalMilliseconds * 0.9);
```

marcin-krystianc (dismissed):
Looks good.

I think it is a good opportunity to add some documentation around LockTryOnce (https://github.com/G-Research/consuldotnet/blob/master/Consul/Lock.cs#L610) option to avoid future confusions:
```
LockTryOnce = false - Acquire method will block forever until the lock is acquired. LockWaitTime is ignored in this case.
LockTryOnce = true - Acquire the lock within a timestamp (It is analogous to `SemaphoreSlim.Wait(Timespan timeout)`. 
                     Under the hood, it attempts to acquire the lock multiple times if needed (due to the HTTP Long Poll returning early), and will do so as many times as it can within the bounds set by LockWaitTime. 
                     If LockWaitTime is set to 0, there will be only single attempt to acquire the lock. 
```

jgiannuzzi (approved):

jgiannuzzi (approved):

PR #27, opened 2020-09-02
Changed Consul.Request to accept multiple values

hovik-aghajanyan:


jgiannuzzi:
Thanks for your PR, @hovik-aghajanyan!

I would like you to split it in multiple smaller PRs, as I see multiple changes mixed together:
1. Use of string interpolation instead of `string.Format`: this is a great modernisation of the codebase. Could you please also convert the instances in `Consul/Coordinate.cs` and `Consul.Test/ClientTest.cs`?
2. Adding `.idea` to `.gitignore`: great to include other IDEs as well. I think that adding `.idea/.idea.Consul/` is a NOOP though.
3. Changing `ConsulRequest.Params` to be a `Dictionary<string, string[]>`: awesome
4. Adding support for filtering with multiple tags: awesome too. I have a few remarks:
   - Could you please also add support for it in `Health.Service`? (see https://www.consul.io/api-docs/health#tag)
   - I don't think we should obsolete the other methods
   - Tests are now failing because the method without tag filtering creates an empty tag param. The issue is at line https://github.com/G-Research/consuldotnet/pull/27/files#diff-d98ef0bc1bea84975e5f1a17036358a9R236, where you should filter out empty or null strings.

I am happy for 3 and 4 to be in the same PR.

jgiannuzzi:
Great work on those additional commits, @hovik-aghajanyan!

As requested above, could you please split this into 3 or 4 new _pull requests_? This will improve the review process and the git history. In the meantime, I will close this PR.

Thanks!

PR #31, opened 2020-09-09
Compatibility with Consul 1.7.x

jgiannuzzi:
This PR fixes a few compatibility issues with Consul 1.7.x and above (latest version is 1.8.3 at the time of writing).

- Consul 1.7 requires an existing node name for Agent.ForceLeave, breaking `Consul.Test.AgentTest.Agent_ForceLeave`
- Consul 1.7 requires a body with a JSON object for Token.Clone, breaking `Consul.Test.TokenTest.Token_ReadCloneUpdateDelete`
    Also added a Clone method that allows to set the description for the new token.
- Consul 1.7 requires proper PolicyLink and RoleLink structures, breaking multiple Role and Token tests
    Made PolicyEntry and RoleEntry base classes instead of subclasses of PolicyLink and RoleLink respectively.
    Added implicit conversion methods from xEntry to xLink for convenience and backwards compatibility.

It also removes the useless test `Consul.Test.PolicyTest.Policy_ReadCloneUpdateDelete`. This test was essentially a copy of `Consul.Test.TokenTest.Token_ReadCloneUpdateDelete`, but trying to delete a nonexistent policy, and not even deleting the cloned token.

An example CI run with Consul 1.8.3 can be found here: https://github.com/jgiannuzzi/consuldotnet/actions/runs/246499369

### DIFF
--- a/Consul/Token.cs
+++ b/Consul/Token.cs
@@ -18,6 +18,7 @@
 
 using Newtonsoft.Json;
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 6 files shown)


Consul/Token.cs
```diff
@@ -224,9 +225,26 @@ public Task<WriteResult<TokenEntry>> Clone(string id, CancellationToken ct = def
         /// <param name="writeOptions">Customized write options</param>
         /// <param name="ct">Cancellation token for long poll request. If set, OperationCanceledException will be thrown if the request is cancelled before completing</param>
         /// <returns>A write result containing the created ACL token</returns>
-        public async Task<WriteResult<TokenEntry>> Clone(string id, WriteOptions writeOptions, CancellationToken ct = default(CancellationToken))
+        public Task<WriteResult<TokenEntry>> Clone(string id, WriteOptions writeOptions, CancellationToken ct = default(CancellationToken))
```
lorinck, line 228: Has the async been deleted deliberatly?

jgiannuzzi: Yes, to be consistent with the way the other methods call their concrete async method.

See for example: https://github.com/G-Research/consuldotnet/blob/4ac6d334b05b27771bcca95b6327d5592d935435/Consul/Token.cs#L163-L185 

The existing `Clone(string id, CancellationToken ct)` method was also like that:
https://github.com/G-Research/consuldotnet/blob/4ac6d334b05b27771bcca95b6327d5592d935435/Consul/Token.cs#L210-L219

lorinck (approved):

PR #32, opened 2020-09-11
Remove [Obsolete] attribute for ConsulClient HttpClient constructor

### LINKED ISSUE #24: Support for injecting HttpClient in a non-obsolete constructor
As of now ConsulClient allows for injecting an HttpClient using an [Obsolete] constructor ([here](https://github.com/G-Research/consuldotnet/blob/master/Consul/Client.cs#L446)). This project should remove this [Obsolete] attribute until an alternative way to inject HttpClient is provided as a migration path (my suggestion is sticking with the current constructor or just a Func\<HttpClient\>). As of now the non-obsolete constructors carry the issue that they internally manage an HttpClient, which means that you either dispose ConsulClient and risk socket exhaustion, or you use a singleton ConsulClient and risk dealing with a stale DNS (see [here](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests)). My personal opinion is that ConsulClient should not be managing HttpClient lifetimes at all since they hide serious issues from the library user.

Salgat:
This is to allow continued support for providing an external HttpClient to the ConsulClient. Closes #24

### DIFF
--- a/Consul/Client.cs
+++ b/Consul/Client.cs
@@ -443,8 +443,6 @@ namespace Consul
         /// </summary>
         /// <param name="config">A Consul client configuration</param>
         /// <param name="client">A custom HttpClient</param>
-        [Obsolete("This constructor is no longer necessary due to the new Action based constructors and will be removed in a future release." +
-            "Please use one of the ConsulClient(Action<>) constructors instead to set internal options on the HttpClient/WebRequestHandler.", false)]
         public ConsulClient(ConsulClientConfiguration config, HttpClient client)
         {
             var ctr = new ConsulClientConfigurationContainer(config, client);
--- a/Consul/Client.cs
+++ b/Consul/Client.cs
@@ -442,7 +442,7 @@ namespace Consul
         /// The HttpClient must accept the "application/json" content type and the Timeout property should be set to at least 15 minutes to allow for blocking queries.
         /// </summary>
         /// <param name="config">A Consul client configuration</param>
-        /// <param name="client">A custom HttpClient</param>
+        /// <param name="client">A custom HttpClient. The lifetime, including disposal, of this HttpClient is not handled by ConsulClient</param>
         public ConsulClient(ConsulClientConfiguration config, HttpClient client)
         {
             var ctr = new ConsulClientConfigurationContainer(config, client);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Added the possibility to specify more http request options and meta-data when registrating service checks. 
   It is now possible to specify the http Header(s) (`Headers`), `Method` and `Body` to be used for a given service check.
   A given service check might also now have an identifier (`ID`), `Name` and a description (`Notes`) associated.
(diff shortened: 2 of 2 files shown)


lorinck (dismissed):
LGTM as discussed under https://github.com/G-Research/consuldotnet/issues/24.
Thank you for raising the PR @Salgat

jgiannuzzi:
@Salgat @lorinck @maurosbarigia I wonder whether we shouldn't also un-obsolete the `ConsulClient(ConsulClientConfiguration config)` constructor?

I understand that @Salgat wants to be able to pass an external `HttpClient` to the constructor, but leaving the other constructor obsolete does not sound consistent. What do you think?

Salgat:
If uniformity is desired it may be more appropriate to leave both obsolete and add a "ConsulClient(Action<ConsulClientConfiguration> configOverride, HttpClient httpClient)" constructor, but I'm good with either way since this is an opinionated decision.

Salgat:
Would it be possible to merge this minor change in for now? This can always be updated later when/if a decision is made on what constructors are ultimately decided on. The goal of this PR is more intent than anything else (since it does not change the behavior of anything, only prevents removing this constructor until a better alternative is present).

Consul/Client.cs
```diff
@@ -443,8 +443,6 @@ public ConsulClient(ConsulClientConfiguration config)
         /// </summary>
         /// <param name="config">A Consul client configuration</param>
         /// <param name="client">A custom HttpClient</param>
```
marcin-krystianc, line 445: It would be nice to add a comment, that it is responsibility of the user to handle the lifecycle of `httpclient`.

marcin-krystianc (commented):
@Salgat apologies for the late response. Before we merge this, can you bump the [version number](https://github.com/G-Research/consuldotnet/blob/master/Consul/Consul.csproj#L4) and add a new entry to the [changelog](https://github.com/G-Research/consuldotnet/blob/master/CHANGELOG.md) ?

jgiannuzzi:
> @Salgat apologies for the late response. Before we merge this, can you bump the [version number](https://github.com/G-Research/consuldotnet/blob/master/Consul/Consul.csproj#L4) and add a new entry to the [changelog](https://github.com/G-Research/consuldotnet/blob/master/CHANGELOG.md) ?

@Salgat In the meantime we changed this guidance, could you update/rebase your branch and simply add an entry to the [changelog](https://github.com/G-Research/consuldotnet/blob/master/CHANGELOG.md)?

Salgat:
Sorry for the delay! I'll see about updating this later today or tomorrow.

Salgat:
Alright should be all good now. Let me know if any other changes are needed.

Consul/Client.cs
Salgat: Great suggestion @marcin-krystianc I updated the documentation for this parameter.

lorinck (approved):

lorinck:
Thanks for raising @Salgat

jgiannuzzi:
Thanks a lot for your contribution to Consul.NET, @Salgat!

PR #34, opened 2020-09-17
Added multiple parameter in request

HAghajanyan-ST:
Added possibility to pass multiple parameters with same name in ConsulRequest

### DIFF
--- a/Consul.Test/ClientTest.cs
+++ b/Consul.Test/ClientTest.cs
@@ -107,10 +107,10 @@ namespace Consul.Test
 
             await Assert.ThrowsAsync<ConsulRequestException>(async () => await request.Execute(CancellationToken.None));
 
-            Assert.Equal("foo", request.Params["dc"]);
+            Assert.Equal("foo", request.Params["dc"].Single());
             Assert.True(request.Params.ContainsKey("consistent"));
-            Assert.Equal("1000", request.Params["index"]);
-            Assert.Equal("1m40s", request.Params["wait"]);
+            Assert.Equal("1000", request.Params["index"].Single());
+            Assert.Equal("1m40s", request.Params["wait"].Single());
         }
 
         [Fact]
@@ -127,8 +127,8 @@ namespace Consul.Test
 
             await Assert.ThrowsAsync<ConsulRequestException>(async () => await request.Execute(CancellationToken.None));
 
-            Assert.Equal("foo", request.Params["dc"]);
-            Assert.Equal("1m40s", request.Params["wait"]);
+            Assert.Equal("foo", request.Params["dc"].Single());
+            Assert.Equal("1m40s", request.Params["wait"].Single());
         }
         [Fact]
         public async Task Client_SetWriteOptions()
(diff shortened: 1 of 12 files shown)


HAghajanyan-ST:
@jgiannuzzi please review PR

jgiannuzzi:
Thanks for your PR, @HAghajanyan-ST!

I will review it soon 👀

jgiannuzzi:
Hi @HAghajanyan-ST, after discussing offline with @marcin-krystianc, we think that we should use [`NameValueCollection`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.namevaluecollection?view=netcore-3.1) for storing the request params, as it is a more direct mapping to what HTTP request params are. Could you please update your PR to use this class instead of `Dictionary<string, IList<string>>`?

I really appreciate the time you have already spent on this so far, and I would understand if you wouldn't want to make that change. In that case, just let us know and we can update your PR ourselves, then let you review it before merging.

HAghajanyan-ST:
Hi @jgiannuzzi , thanks for your comment, I didin't use NameValueCollection because, If the specified key already exists in the target NameValueCollection instance, the specified value is added to the existing comma-separated list of values in the form "value1,value2,value3". So in this case we must also check if there is a comma separate the value by comma. If you still think I must change I can do it :) 

Source:
https://stackoverflow.com/questions/21590778/add-duplicates-keys-to-namevaluecollection

marcin-krystianc:
Hi @HAghajanyan-ST,

Thanks for your response. I think that the msdn documentation is misleading 🤷.
`NameValueCaollection` has two ways of getting values: `Get` and `GetValues`.
What you say is correct only for the `Get` method, as it returns a single string with all values joined with the `,`.
However the `GetValues` returns an array of strings, so it is not a problem even if any of these values contains a `,` character.

E.g.:
```
var nameValueCollection = new NameValueCollection();
nameValueCollection.Add("a", "Hello,Hello");
nameValueCollection.Add("a", "World,World");

Console.WriteLine(nameValueCollection.Get("a"));
foreach (var v in nameValueCollection.GetValues("a"))
{
    Console.WriteLine(v);
}
```
```
Hello,Hello,World,World
Hello,Hello
World,World
```

jgiannuzzi:
Hi @HAghajanyan-ST, as @marcin-krystianc explained above, we think that NameValueCollection is the right way to go.
Would you mind adapting your PR to use it, or should we?

mfkl:
Some initial work here. https://github.com/mfkl/consuldotnet/commits/feature/multiple-tags

Rebased and started using `NameValueCollection` instead. It's not ready yet as it appears to be failing some tests, and I'm not too sure on how to go about the `BuildConsulUri` new implementation.

mfkl:
> Do you still think that this change is necessary after the https://github.com/G-Research/consuldotnet/pull/36 was implemented?

Yes, I believe indeed #36 should do it. Feel free to let us know if that works for you @HAghajanyan-ST.

Closing.

PR #35, opened 2020-09-20
Added Rider config folder in gitignore

HAghajanyan-ST:

PR #36, opened 2020-09-22
Search service by multiple tags

HAghajanyan-ST:


mfkl:
You can revert https://github.com/G-Research/consuldotnet/pull/36/commits/445ba5a8b2edfae35637d9ae9eaecbef2ee8ff6a from this branch, thanks to https://github.com/G-Research/consuldotnet/pull/35.

Personally not a fan of merge commits, rebase is cleaner, but that's just me.

Consul.Test/ClientTest.cs
```diff
@@ -127,8 +127,8 @@ public async Task Client_SetClientOptions()
 
             await Assert.ThrowsAsync<ConsulRequestException>(async () => await request.Execute(CancellationToken.None));
 
-            Assert.Equal("foo", request.Params["dc"]);
-            Assert.Equal("1m40s", request.Params["wait"]);
+            Assert.Equal("foo", request.Params["dc"].Single());
```
lorinck, line 130: Should there be a test which is actually working with multiple values?

Consul/Catalog.cs
```diff
@@ -211,12 +212,26 @@ public Task<QueryResult<CatalogService[]>> Service(string service, string tag, C
...
         public Task<QueryResult<CatalogService[]>> Service(string service, string tag, QueryOptions q, CancellationToken ct)
+        {
+            return Service(service, new[] {tag}, q, ct);
+        }
+        
+        /// <summary>
+        /// Service is used to query catalog entries for a given service
+        /// </summary>
+        /// <param name="service">The service ID</param>
+        /// <param name="tags">A tag to filter on</param>
```
lorinck, line 223: The text should say 'Tags' not 'A tag'

Consul/Interfaces/IHealthEndpoint.cs
```diff
@@ -35,6 +36,7 @@ public interface IHealthEndpoint
         Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, CancellationToken ct = default(CancellationToken));
         Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, bool passingOnly, CancellationToken ct = default(CancellationToken));
         Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, bool passingOnly, QueryOptions q, CancellationToken ct = default(CancellationToken));
+        Task<QueryResult<ServiceEntry[]>> Service(string service, IList<string> tag, bool passingOnly, QueryOptions q, CancellationToken ct = default(CancellationToken));
```
lorinck, line 39: This should be called tags now its a collection.

Consul/Catalog.cs
```diff
@@ -211,12 +212,26 @@ public Task<QueryResult<CatalogService[]>> Service(string service, string tag, C
...
+        /// <param name="q">Customized query options</param>
+        /// <param name="ct">Cancellation token for long poll request. If set, OperationCanceledException will be thrown if the request is cancelled before completing</param>
+        /// <returns>A list of service instances</returns>
+        public Task<QueryResult<CatalogService[]>> Service(string service, IList<string> tags, QueryOptions q, CancellationToken ct)
         {
             var req = _client.Get<CatalogService[]>(string.Format("/v1/catalog/service/{0}", service), q);
-            if (!string.IsNullOrEmpty(tag))
+            if (tags?.Any(s => !string.IsNullOrEmpty(s)) == true)
```
lorinck, line 230: I think it would be nice to drop empty tags here, not adding everything to the request.

Consul/Client_Request.cs
```diff
@@ -91,10 +92,9 @@ protected Uri BuildConsulUri(string url, Dictionary<string, string> p)
             var queryParams = new List<string>(Params.Count / 2);
             foreach (var queryParam in Params)
             {
-                if (!string.IsNullOrEmpty(queryParam.Value))
+                if (queryParam.Value?.Any() == true)
```
lorinck, line 95: This doesn't look right. Empty values will be still added <key>=''

lorinck (changes requested):
It would be nice to actually test the new logic of having multiple values allowed rather than defaulting everywhere to read to first one from the collection to make the old tests pass.

PR #39, opened 2020-10-16
Allow further service check property to be specified when registering a service

gboder:
Added additional Consul HTTP API parameter to detail the service check registration.

Reference to <https://www.consul.io/docs/discovery/services> and this <https://www.consul.io/docs/discovery/checks> page of Consul's documentation.

We are going to use those options to allow specify distinct host http header and host address values.

Sample definition
```csharp
new AgentServiceCheck{
	HTTP = $"http://srv1.servers.contoso.com/weatherforecast",
	Timeout = TimeSpan.FromMilliseconds(100),
	Interval = TimeSpan.FromSeconds(1),
	Header = new Dictionary<string, List<string>>{ {"Host",new List<string>{"app.contoso.com"} } },
	Method = "GET",
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,12 @@
 # Changelog
 
-## 1.6.1.3
-* Added the possibility to specify the id, name and notes for a service check definition.
-  Added the possibility to indicate when a dead service should be dropped from the registry `DeregisterCriticalServiceAfter` http API parameter
+## 1.7.0
+* Added the possibility to specify more http request options and meta-data when registrating service checks. 
+  It is now possible to specify the http Header(s) (`Headers`), `Method` and `Body` to be used for a given service check.
+  A given service check might also now have an identifier (`ID`), `Name` and a description (`Notes`) associated.
+  Details on the Consul's HTTP API can be found here <https://www.consul.io/api/agent/check>
 
-## 1.6.1.2
-* Added the possibility to specify more http request options when registrating service checks.
-  It is now possible to specify the Header(s), method and body to be used for a given service check.
-  Details might be found here <https://www.consul.io/api/agent/check>
+* Increased minor version and align with NuGet versioning best-practices <https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics>
 
 ## 1.6.1.1
 * Fix issue #9 preventing use of the library with .NET Framework
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.6.1.2
+* Added possibility to specify more http request options when registrating service checks.
+  It is now possible to specify the Header(s), method and body to be used for a given service check.
+  Details might be found here <https://www.consul.io/api/agent/check>
+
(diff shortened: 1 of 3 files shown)


gboder:
I'll give a look at the tests' result...

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 1.6.1.3
```
marcin-krystianc, line 3: I'm not sure why did you decide to include two versions in the changelog ?
With your PR `v1.6.1.2` would never be released, so there is no point including it in the changelog.

Consul/Consul.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <VersionPrefix>1.6.1.1</VersionPrefix>
+    <VersionPrefix>1.6.1.3</VersionPrefix>
```
marcin-krystianc, line 4: I would suggest using version `1.7.0`:
- Three-part version number is more common versioning scheme for NuGet, https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics
- According to the semantic versioning rules (https://semver.org/spec/v2.0.0.html) you should bump the `Minor` version:
```
...
- Bug fixes not affecting the API increment the patch version,
- Backwards compatible API additions/changes increment the minor version
- Backwards incompatible API changes increment the major version.
...
```

Consul/Agent.cs
```diff
@@ -227,6 +218,16 @@ public class AgentCheckRegistration : AgentServiceCheck
...
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public string ID { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public string Name { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public string Notes { get; set; }
+
```
marcin-krystianc, line 230: 👍

marcin-krystianc (commented):
Thanks @gboder for making this change.
Changes are good, just left a comment regarding the changelog and versioning.

CHANGELOG.md
gboder: Right - dropping this entry and merging with the other changes. Also adjusting change log towards version '1.7.0'.

Consul/Consul.csproj
gboder: I was also wondering a bit about the 4-digits notation, I only wanted to stay on the same pattern. Now if we think about three digits its for sure a `1.7.0`.

Let's follow the NuGet paradigm. I'll update to `1.7.0`

marcin-krystianc (approved):

marcin-krystianc:
Hi @gboder ,

I reverted your version change back to the `v1.6.1.1`. I didn't realise that we have a rule to keep the version of `Consul.NET` to match the supported API version of `Consul' . Now the information about this is added to the [README](https://github.com/G-Research/consuldotnet/blob/master/README.md#versioning).

PR #38, opened 2020-10-06
Add Consul.AspNetCore CI, refactor project, add options, remove obsol…

sergeyshaykhullin:
Add Consul.AspNetCore CI, refactor project, add options, remove obsolete on client ctor, bump version

- Package Consul.AspNetCore is owned by this implementation https://www.nuget.org/packages/Consul.AspNetCore/
- Should we try to contact to zlzforever or use another package name?

**Decisions**
- [x] [**Discussable**] I added separate workflow, because of ci.yml is required to push and all checks have to be passed
- [x] [**Discussable**] `Consul.AspNetCore` workflow called consul.aspnetcore.yaml
- [x] I removed `Obsolete` from `ConsulClient` ctor to integrate with `IOptions`
- [x] I created `IConsulClientFactory` to integrate with `IOptions` and separate from registration logic. Also to use `NamedOptions` and create override with `CreateClient(string name)` to support multiple clients in one process
- [x] [**Discussable**] I used `Consul | AspNetCore | Package | Artifact | Download` names in workflow steps for readability
- [x] [**Discussable**] I used `3.1.8` version like an AspNetCore, should it be a Consul version? Or independant one?
- [x] [**Discussable**] I separated dists to Consul and Consul.AspNetCore to not to pack both .nupkg in artifact. So dist/consul for Consul and dist/consul.aspnetcore for Consul.AspNetCore
- [x] Tests for Consul.AspNetCore runs on windows-latest only, because of no external os specific dependency required and already written script to extract package version

PR #46, opened 2020-12-04
Add AspNetCore CI

sergeyshaykhullin:
https://github.com/G-Research/consuldotnet/pull/38

### DIFF
--- a/.github/workflows/consul.aspnetcore.yml
+++ b/.github/workflows/consul.aspnetcore.yml
@@ -1,0 +1,64 @@
+name: Consul | AspNetCore
+
+on:
+  push:
+  pull_request:
+
+jobs:
+
+  build:
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
+    name: Consul | AspNetCore
+    runs-on: windows-latest
+
+    steps:
+      - name: Consul | AspNetCore | Checkout
+        uses: actions/checkout@v2
+
+      - name: Consul | AspNetCore | .NET Core
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 3.1.300
+
+      - name: Consul | AspNetCore | Build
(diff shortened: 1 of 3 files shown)


.github/workflows/consul.aspnetcore.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - name: Consul | AspNetCore | Package
+        run: |
+          if ("${{ github.ref }}" -like "refs/tags/v*") {
+            $tag = "${{ github.ref }}".SubString(11)
+            $version = (Select-Xml -path Consul.AspNetCore/Consul.AspNetCore.csproj -XPath "/Project/PropertyGroup/VersionPrefix/text()").node.Value
+            if (-not ($tag -eq $version)) {
+              echo "There is a mismatch between the project version ($version) and the tag ($tag)"
+              exit 1
+            }
```
marcin-krystianc, line 40: We cannot use the same tag name pattern for both packages (`Consul` and `Consul.AspNetCore`) as the version validation logic would be broken. Either  `Consul` or `Consul.AspNetCore` would always have "wrong" version number.

I think we should use a more specific prefix here, e.g. `refs/tags/Consul.AspNetCore_v*`

.github/workflows/consul.aspnetcore.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Consul | AspNetCore | Package | Artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: nuget-aspnetcore-package
+          path: dist/consul.aspnetcore
+
+  # Publish NuGet package when a tag is pushed.
+  # Tests need to succeed first, including having a tag name that matches the version number.
+  publish:
+    if: startsWith(github.ref, 'refs/tags/v')
```
marcin-krystianc, line 55:
```suggestion
    if: startsWith(github.ref, 'refs/tags/Consul.AspNetCore_v')
```

marcin-krystianc (approved):
LGTM 👍

PR #55, opened 2020-12-10
Use fewer concurrent http connections for testing

marcin-krystianc:
- The number of threads for running the test suite is reduced to 20. This is to avoid using too many concurrent http connections with consul agent
- Consul agent has a config option `limit.http_max_conns_per_client ` which could be set to reduce number of accepted concurrent http connections from single client. We don't set it only because it is not supported in v1.6.1 which is currently used in our tests.


jgiannuzzi (approved):
LGTM 👍

Thank you @marcin-krystianc, this will help us run the CI with Consul > 1.6.3!

PR #56, opened 2020-12-10
Avoid running too many tests in parallel

marcin-krystianc:
Each test creates one or more http connections to the Consul agent. 
Before this change all tests were run in parallel, so the peak number http connections was reaching over one hundred.
This change reduces the number of http connections that are required to complete all tests.
To control the parallelism of tests we use `xUnit` functionality (https://xunit.net/docs/running-tests-in-parallel).


Consul.Test/AgentTest.cs
```diff
@@ -479,18 +476,32 @@ public async Task Agent_NodeMaintenance()
         [Fact]
         public async Task Agent_Monitor()
         {
-            var logs = await _client.Agent.Monitor(LogLevel.Trace);
-            var counter = 0;
-            var successToken = new CancellationTokenSource();
-            var failTask = Task.Delay(5000, successToken.Token).ContinueWith(x => Assert.True(false, "Failed to finish reading logs in time"));
-            foreach (var line in logs)
+            using (var logs = await _client.Agent.Monitor(LogLevel.Trace))
```
marcin-krystianc, line 479: Logs are `IDisposable` so me need to wrap it with `using ` statement.

Consul.Test/ClientTest.cs
```diff
@@ -117,19 +114,22 @@ public async Task Client_SetQueryOptions()
         [Fact]
         public async Task Client_SetClientOptions()
         {
-            var client = new ConsulClient(c =>
+            using (var client = new ConsulClient(c =>
```
marcin-krystianc, line 117: So the client gets properly disposed.

Consul.Test/Properties/AssemblyInfo.cs
```diff
@@ -34,5 +34,9 @@
 [assembly: AssemblyVersion("1.0.0.0")]
 [assembly: AssemblyFileVersion("1.0.0.0")]
 
-// This is to avoid using too many concurrent http connections with consul agent
-[assembly: CollectionBehavior(MaxParallelThreads = 20)]
```
marcin-krystianc, line 38: `MaxParallelThreads` doesn't actually reduce the number of tests that can be run in parallel. It only sets the limit on the number of threads in the thread pool (which doesn't really affect parallelism of asynchronous code).

Consul.Test/AgentTest.cs
```diff
@@ -479,18 +476,32 @@ public async Task Agent_NodeMaintenance()
...
+                        counter++;
+                        if (counter > 5)
+                        {
+                            successToken.Cancel();
+                            break;
+                        }
+                    }
+                });
+
+                await await Task.WhenAny(new[] { timeoutTask, logsTask });
```
jgiannuzzi, line 503: Is `await await` on purpose here? (this is not a rhetorical question!)

jgiannuzzi (commented):
This is great, thanks @marcin-krystianc!

I really appreciate that this change does not slow down the complete test run 😄

I just have one little comment below about something that _looks_ like a typo to me, but might be on purpose.

mfkl (approved):

Consul.Test/AgentTest.cs
marcin-krystianc: No it is not a typo 😄 
- `Task.WhenAny` creates a new task that returns one of the provided tasks
- first `await` gives as the task that completed first
- second `await` awaits the completed task which either does nothing or throws an exception (if the task has failed with an exception) 


Previously that code didn't work at all, because the exception from timeout assertion was never propagated to the main thread.

jgiannuzzi (approved):
LGTM 👍
Thanks @marcin-krystianc!

PR #57, opened 2020-12-15
Upload logo files in various formats

c-rindi:
Uploading all the logo files recently made, in several formats: .ai, .jpg, .png, .svg.


mfkl:
Cool 👍 

I suggest creating an `asset` folder for these, to keep the root of the repository lean :-)

I'm guessing the plan is to add one of these logos to the README?

c-rindi:
Thanks @mfkl. 
Yeah, the plan is to use the most colorful one for the read.me file. 
It took me this long to figure out how to create a folder for them though so who knows how long it will take for me to resize it appropriately and add it. LOL.

stackedsax:
Hey there @mkfl.  Looks like there's already an assets directory in here, but it has a key file in it so it doesn't feel like the same kind of "assets" as these branding docs.  Is the "Logo" directory a fine enough place for these logo files for now, or should we create a "Docs" folder in anticipation of more types of documents related to consuldotnet?

Also, I assume that Capitalizing the first letter of filenames/directories in dotnet projects is the norm?  That's what it looks like in here, just wanted to check.

mfkl:
> Looks like there's already an assets directory in here, but it has a key file in it so it doesn't feel like the same kind of "assets" as these branding docs.

Indeed, didn't notice!

> Is the "Logo" directory a fine enough place for these logo files for now, or should we create a "Docs" folder in anticipation of more types of documents related to consuldotnet?

Both are fine, I guess. Unsure whether we will have many docs documents in the future, the upstream consul API docs might be enough?

> Also, I assume that Capitalizing the first letter of filenames/directories in dotnet projects is the norm?

I don't think there is a clear guidance on this. Browsing https://github.com/dotnet/runtime/ shows that it uses lowercase for most directories, except when the directory is part of a dotnet namespace, then it follows Pascal case.

jgiannuzzi:
CI tests failure issue is tracked in #60. Once solved, we should be able to merge this PR.

stackedsax (approved):
👍

PR #59, opened 2021-01-21
Upload Consul logs in case of test failure, and hide them by default

jgiannuzzi:
This PR improves the CI by hiding the Consul logs, which helps reading the test results, and uploading them as an artifact in case of failure.


marcin-krystianc (dismissed):
LGTM

.github/workflows/ci.yml
```diff
@@ -69,8 +69,14 @@ jobs:
         shell: bash
         run: |
           cd Consul.Test
-          ./consul agent -dev -config-file test_config.json &
+          ./consul agent -dev -config-file test_config.json -log-file consul.log >/dev/null &
```
marcin-krystianc, line 72: Do we need to redirect `stderr` as well ?

jgiannuzzi: Good catch, I will redirect `stdout` and `stderr` to different files — in case of failures, we'd rather more information than less.

marcin-krystianc: I think it would be even better to redirect them into the same file, e.g.: https://askubuntu.com/questions/625224/how-to-redirect-stderr-to-a-file

marcin-krystianc (approved):

.github/workflows/ci.yml
jgiannuzzi: I thought about it, but figured that it would make failure investigations easier to have them separate. What do you think?

marcin-krystianc: I don't have strong opinion here. Let's try as it is and see if it is helpful.

jgiannuzzi:
CI tests failure issue is tracked in #60. Once solved, we should be able to merge this PR.

PR #62, opened 2021-01-21
Fix the flakiness of Agent_Monitor test

marcin-krystianc:
By default the `ContinueWith` runs the continuation regardless of the reason for completing the preceding task, whether it completed normally or was cancelled via cancellation token. As a result even when `logsTask` completed sooner, the `timeoutTask` was completing immediately after it due to usage of cancellation token. It was a race condition, because `Task.WhenAny` could select either task.

After this change there is no race condition anymore.


jgiannuzzi (approved):
That's much better 👍

Thanks, @marcin-krystianc!

PR #63, opened 2021-01-21
Fix the flakiness of the Event_FireList test by using blocking queries

marcin-krystianc:
Fixes the flakiness of the `Event_FireList` by usage of Consul blocking queries feature.

### DIFF
--- a/Consul.Test/EventTest.cs
+++ b/Consul.Test/EventTest.cs
@@ -18,6 +18,7 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


Consul.Test/EventTest.cs
```diff
@@ -45,21 +45,15 @@ public void Dispose()
...
-            await Task.Delay(100);
-
+            var previousIndex = (await _client.Event.List()).LastIndex;
+            var userEvent = new UserEvent() {Name = "foo"};
+            var res = await _client.Event.Fire(userEvent);
             Assert.NotEqual(TimeSpan.Zero, res.RequestTime);
             Assert.False(string.IsNullOrEmpty(res.Response));
 
-            var events = await _client.Event.List();
+            var events = await _client.Event.List("foo", new QueryOptions {WaitIndex = previousIndex});
```
jgiannuzzi, line 54: I would use `userEvent.Name` instead of `"foo"` here

Consul.Test/EventTest.cs
```diff
@@ -45,21 +45,15 @@ public void Dispose()
...
         public async Task Event_FireList()
         {
-            var userevent = new UserEvent()
-            {
-                Name = "foo"
-            };
-
-            var res = await _client.Event.Fire(userevent);
-
-            await Task.Delay(100);
```
marcin-krystianc, line 55: We don't need any delay on client side because we use `WaitIndex` (https://www.consul.io/api/features/blocking).

jgiannuzzi (commented):
I _think_ this should be wrapped in `TimeoutUtils.WithTimeout` so we don't wait forever in case of issues.

marcin-krystianc:
> _think_ this should be wrapped in `TimeoutUtils.WithTimeout` so we don't wait forever in case of issues.
> 
> Edit: though maybe using `WaitTime` in the query options would be enough? Maybe we should do both?

If `WaitTime` is not set then the default timeout is used which according to my tests is 10 min (it is definitely not forever 😄 ).
I think it is better to avoid adding any extra timeout logic (like `TimeoutUtils.WithTimeout`) if it is not absolutely necessary, I believe it is good as it is now.

jgiannuzzi (approved):
Alright, approved then!

Thanks @marcin-krystianc 👍

PR #64, opened 2021-01-21
Fix lock and semaphore timeouts

marcin-krystianc:
The logic for calculating the `WaitTime` was broken.
I think that this bug was inherited from go library. It is fixed there already (https://github.com/hashicorp/consul/commit/9043966efdab6e8e12c73772de070e25c4a50bc1)

### DIFF
--- a/Consul/Lock.cs
+++ b/Consul/Lock.cs
@@ -252,7 +252,7 @@ namespace Consul
 
                     var qOpts = new QueryOptions()
                     {
-                        WaitTime = Opts.LockWaitTime,
+                        WaitTime = Opts.LockWaitTime
                     };
 
                     var attempts = 0;
--- a/Consul/Lock.cs
+++ b/Consul/Lock.cs
@@ -17,6 +17,7 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Diagnostics;
 using System.Runtime.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


Consul/Lock.cs
```diff
@@ -251,23 +252,24 @@ public async Task<CancellationToken> Acquire(CancellationToken ct)
 
                     var qOpts = new QueryOptions()
                     {
-                        WaitTime = Opts.LockWaitTime
+                        WaitTime = Opts.LockWaitTime,
                     };
 
                     var attempts = 0;
-                    var start = DateTime.UtcNow;
```
marcin-krystianc, line 258: `DateTime` arithmetic is not a good way of calculating elapsed time, `StopWatch` is more accurate.

Consul/Lock.cs
```diff
@@ -251,23 +252,24 @@ public async Task<CancellationToken> Acquire(CancellationToken ct)
 
                     var qOpts = new QueryOptions()
                     {
-                        WaitTime = Opts.LockWaitTime
+                        WaitTime = Opts.LockWaitTime,
```
jgiannuzzi, line 255: Please remove the comma at the end for style consistency with the rest of the project

jgiannuzzi (commented):
Great catch, @marcin-krystianc!

I just have one tiny nitpick below, but otherwise it LGTM 👍

jgiannuzzi:
A more in depth explanation of @marcin-krystianc's fix can be found in the [original PR for Consul](https://github.com/hashicorp/consul/pull/4346).

Consul/Lock.cs
marcin-krystianc: 👍

Consul/Lock.cs
```diff
@@ -251,23 +252,24 @@ public async Task<CancellationToken> Acquire(CancellationToken ct)
...
-                            var elapsed = DateTime.UtcNow.Subtract(start);
-                            if (elapsed > qOpts.WaitTime)
+                            var elapsed = sw.Elapsed;
+                            if (elapsed > Opts.LockWaitTime)
                             {
                                 DisposeCancellationTokenSource();
                                 throw new LockMaxAttemptsReachedException("LockTryOnce is set and the lock is already held or lock delay is in effect");
                             }
-                            qOpts.WaitTime -= elapsed;
+
```
jgiannuzzi, line 271: Can you remove the extra line?

jgiannuzzi (approved):
👍

Issue #65, opened 2021-01-25
Lock Acquire operation null reference exception + cancellation token not respected


georgy93:
Hello!

Recently I've been playing with the distributed lock feature of Consul in an ASP Net Core 3.1 application.
I'm using the latest version of consuldotnet 1.6.1.1
I have a Background service that calls the Task<CancellationToken> Acquire(CancellationToken ct) continuously in a "block indefinitely" approach (LockTryOnce is left with its default value).  I.e. If I am not a leader, then wait until I am. If I was previously a leader, try become leader again in case I lost the leadership (e.g session invalidation). 

However, I have the following two issues :
- If I call Acquire method with an already cancelled token I get a Null Reference Exception from whitin the consul library. 
- If the lock is already held by another instace of my application and my current "slave" instance is in a blocked Acquire call, then when I perform a gracefull shutdown, my BackgroundService cancellaiton token is not respected although I'm passing it to the Acquire method. The Acqurie method still keeps the blocked operation until "LockWaitTime" has passed. Then I again get a Null Reference exception. 
What is even worse is that in order to have my gracefull shutdown complete normally for the entire application I must either increase the ASP Net Core Host ShutDownTimeout or have some custom extension method where I for example do:
"await Task.WhenAny(AcquireLockTask, TaskThatCompletesUpponCancellationTokenCancelled);"

Edit:
I've just done a quick debug session and I found that:
In the finally block of the Acquire method If I have never started a monitor task (e.g my application never acquried a lock, during its life) it will throw the null reference exception because:
 "await _monitorTask.ConfigureAwait(false);" will be "await null.ConfigureAwait(false)".
https://github.com/G-Research/consuldotnet/blob/master/Consul/Lock.cs#L357

 However adding an "if" to check if the task _monitorTask is initialised would lead to 
 "throw new LockNotHeldException("Unable to acquire the lock with Consul");" 
https://github.com/G-Research/consuldotnet/blob/master/Consul/Lock.cs#L345
because this was the last thing from the method before the finally block. I'm not sure if this exception would be correct in case of cancellation token being cancelled. Could be, could be not.

For the second issue where the CancellationToken is not respected:
https://github.com/G-Research/consuldotnet/blob/master/Consul/Lock.cs#L278 
https://github.com/G-Research/consuldotnet/blob/master/Consul/Lock.cs#L330
[...]

PR #66, opened 2021-02-04
Lock: Add missing null check

mfkl:
Fixes part of https://github.com/G-Research/consuldotnet/issues/65 (should be split in 2 issues).

This resulted in a NullReferenceException hiding the expected (?) LockNotHeldException.

### DIFF
--- a/Consul.Test/SemaphoreTest.cs
+++ b/Consul.Test/SemaphoreTest.cs
@@ -32,6 +32,8 @@ namespace Consul.Test
     public class SemaphoreTest : IDisposable
     {
         private ConsulClient _client;
+        const int DefaultSessionTTLSeconds = 10;
+        const int LockWaitTimeSeconds = 15;
 
         public SemaphoreTest()
         {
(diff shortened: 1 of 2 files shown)


mfkl:
@georgy93 mentions 

> `LockNotHeldException` because this was the last thing from the method before the finally block. I'm not sure if this exception would be correct in case of cancellation token being cancelled. Could be, could be not.

Consul.Test/SemaphoreTest.cs
```diff
@@ -363,5 +365,61 @@ public async Task Semaphore_Conflict()
...
+                SessionTTL = TimeSpan.FromSeconds(DefaultSessionTTLSeconds),
+                LockWaitTime = TimeSpan.FromSeconds(LockWaitTimeSeconds)
+            });
+
+            var cts = new CancellationTokenSource(TimeSpan.FromSeconds(1));
+
+            await distributedLock2.Acquire(); // Become "Master" with another instance first
+            var task = distributedLock.Acquire(cts.Token);
+
+            await Assert.ThrowsAsync<LockNotHeldException>(async () => await distributedLock.Acquire(cts.Token));
```
marcin-krystianc, line 415: `distributedLock.Acquire(cts.Token)` is now called twice - I don't think you need the ```var task = distributedLock.Acquire(cts.Token);```

Consul.Test/SemaphoreTest.cs
```diff
@@ -363,5 +365,61 @@ public async Task Semaphore_Conflict()
...
+
+            await distributedLock2.Release();
+            if (distributedLock.IsHeld)
+                await distributedLock2.Release();
```
marcin-krystianc, line 419: One lock is used in the condition, but the other one is released - is it correct ?

Consul.Test/SemaphoreTest.cs
```diff
@@ -363,5 +365,61 @@ public async Task Semaphore_Conflict()
...
+            await distributedLock2.Destroy();
```
marcin-krystianc, line 420: `distributedLock2` is not being destroyed, I think you should destroy it as well.

marcin-krystianc (commented):
Few remarks regarding the test code.

Consul.Test/SemaphoreTest.cs
mfkl: done

Consul.Test/SemaphoreTest.cs
mfkl: fixed

Consul.Test/SemaphoreTest.cs
mfkl: with the up to date version of the code, only `distributedLock2` is being destroyed as calling destroy on `distributedLock` doesn't do anything, because it was never acquired in the first place.

mfkl:
Hi @georgy93, could you please review this PR and confirm this is your actual use case?

Consul.Test/SemaphoreTest.cs
```diff
@@ -363,5 +365,61 @@ public async Task Semaphore_Conflict()
...
         }
+
+        [Fact]
+        public async Task Passing_Cancelled_CancellationToken_Should_Throw_LockNotHeldException()
+        {
+            const string keyName = "service/myApp/leader";
+
+            var distributedLock = _client.CreateLock(new LockOptions(keyName)
+            {
+                SessionTTL = TimeSpan.FromSeconds(DefaultSessionTTLSeconds),
```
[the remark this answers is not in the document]
mfkl: Fine by me either way. https://github.com/G-Research/consuldotnet/pull/67 uses them too

marcin-krystianc (approved):

georgy93:
Hi @mfkl !
I played around with the changes and they seem okay to me. 
No more null reference exceptions and when I cancel the cancellation token I get TaskCancelledException and the operation stops as expected.

mfkl:
Let's merge this? @marcin-krystianc @jgiannuzzi

Consul.Test/SemaphoreTest.cs
```diff
@@ -363,5 +365,57 @@ public async Task Semaphore_Conflict()
...
+                LockWaitTime = TimeSpan.FromSeconds(LockWaitTimeSeconds)
+            });
+
+            var cts = new CancellationTokenSource(TimeSpan.FromSeconds(1));
+
+            await distributedLock2.Acquire(); // Become "Master" with another instance first
+
+            await Assert.ThrowsAsync<LockNotHeldException>(async () => await distributedLock.Acquire(cts.Token));
```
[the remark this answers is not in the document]
mfkl: Done, thanks.

Consul.Test/SemaphoreTest.cs
```diff
@@ -363,5 +365,57 @@ public async Task Semaphore_Conflict()
...
+
+            cts.Cancel();
+
+            await Assert.ThrowsAsync<LockNotHeldException>(async () => await distributedLock.Acquire(cts.Token));
+        }
+
+        [Fact]
+        public async Task Cancelling_A_Token_When_Acquiring_A_Lock_Should_Throw_LockNotHeldException()
+        {
+            const string keyName = "service/myApp/leader";
```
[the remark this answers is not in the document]
mfkl: Done

Issue #69, opened 2021-02-12
Failure to run test on Windows with net461 TFM


mfkl:
**Describe the bug**

There is a problem when running Consul.Test on Windows with net461 TFM. Fine with netcoreapp2.1 (which the CI uses).

**Steps To Reproduce**

Just run the test in VS or CLI on master.

**Expected behavior**

Tests pass just like with the net core TFM.

**Logs**

`FileLoadException` for `Newtonsoft.Json` for almost all tests.

Example:
```
Message: 
    System.IO.FileLoadException : Could not load file or assembly 'Newtonsoft.Json, Version=10.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
  Stack Trace: 
    ConsulClient.ctor(Action`1 configOverride, Action`1 clientOverride, Action`1 handlerOverride)
    ConsulClient.ctor(Action`1 configOverride) line 384
    ACLTest.ctor() line 33
```

**Environment**

- OS:       Windows
- Consul Version:   Latest
- consultdotnet Version   Latest

PR #71, opened 2021-02-19
Tests: Add Newtonsoft.Json reference to fix test run for net461

mfkl:
Considered using https://github.com/ravibpatel/ILRepack.Lib.MSBuild.Task instead but it'd mean we count on it being maintained in the future, and we'd need to handle binding redirects.

This is a workaround until we remove the ILRepack step altogether.

Closes https://github.com/G-Research/consuldotnet/issues/69


marcin-krystianc (approved):
LGTM

jgiannuzzi:
Thanks for your PR, @mfkl!

I actually prefer the alternative approach from #72, as it would allow everyone to build and run the tests in the same way as we do in the CI and the released NuGet.

jgiannuzzi:
Closed PR because the alternative approach from #72 got merged.

PR #73, opened 2021-03-01
Convert to System.Text.Json

### LINKED ISSUE #43: Is it possible to remove the dependecy on Newtonsoft.Json?
Hello,

I am using your library to register my services for my API gateway without issues, but I noticed that it still has a reference to Newtonsoft.Json as the only library I use.
Would it be possible to switch to System.Text.Json serializers?

Regards

highlyunavailable:
Convert all uses of Newtonsoft.Json to System.Text.Json

Closes #43

This converts the library to use System.Text.Json instead of Newtonsoft.Json to remove the dependency on that. All tests pass locally, but this is such a fundamental change that there may be unseen issues. This may also cause another problem: would System.Text.Json need to be ILMerged like Newtonsoft.Json? The original reason that it was being ilmerged was to stop dependency issues between common libraries, but I'm not sure if .net still has this problem with the new JSON libs.

There is also a breaking change in here: All references to `dynamic` are gone (it was used in the agent `Self` API) as compilation was failing due to needing the Microsoft.CSharp library referenced. Instead, raw `JsonElement`s are being returned instead where `dynamic` was being used.

### DIFF
--- a/Consul/Interfaces/IAgentEndpoint.cs
+++ b/Consul/Interfaces/IAgentEndpoint.cs
@@ -19,6 +19,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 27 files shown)


Consul/Interfaces/IAgentEndpoint.cs
```diff
@@ -44,7 +45,7 @@ public interface IAgentEndpoint
         string NodeName { get; }
         Task<string> GetNodeName(CancellationToken ct = default(CancellationToken));
         Task PassTTL(string checkID, string note, CancellationToken ct = default(CancellationToken));
-        Task<QueryResult<Dictionary<string, Dictionary<string, dynamic>>>> Self(CancellationToken ct = default(CancellationToken));
+        Task<QueryResult<Dictionary<string, Dictionary<string, JsonElement>>>> Self(CancellationToken ct = default(CancellationToken));
```
marcin-krystianc, line 48: We need to be really careful with changing the public API to avoid runtime failures when [diamond dependency issue](https://codeblog.jonskeet.uk/2019/06/30/versioning-limitations-in-net/) occurs. 
I think that in this case the best would be to keep using the `dynamic` "type". To make it compile we could simply add an explicit package reference to `Microsoft.CSharp Version="4.3.0"`. It will not by any real dependency addition as this package is currently referenced transitively via `Newtonsoft.Json`.

marcin-krystianc (commented):
@highlyunavailable Thank you for submitting this PR, it's excellent work.

Please see my comment on the version of `System.Text.Json` used in this PR. It worries me that its current adoption is too low which can cause troubles for our consumers. I think we should wait for wider adoption of the version `5.0.1` or higher before merging this PR.

PR #70, opened 2021-02-15
Added service and node filtering abilities into QueryOptions

declard:
Added [server-side filtering support](https://www.consul.io/api-docs/features/filtering) as an eDSL

This implementation supports filtering by the following components:
- Service
  - ID
  - Tags
  - Meta

With the following operations:
- and
- or
- not
- ==
- !=
- contains
- is empty

For Health.Service

whenever they are applicable to the listed components

Functionally satisfies #26

PR #74, opened 2021-03-26
Fix recent CI issues

jgiannuzzi:
This PR ~~deletes the obsolete `NuGet.config` file that was adding the `dotnet-core` feed from myget.org~~ removes the obsolete `dotnet-core` feed from myget.org from the `NuGet.config` file.

This fixes recent CI issues like this one: https://github.com/G-Research/consuldotnet/pull/70/checks?check_run_id=2203971524.

~~It also resets the NuGet sources on Windows runners to workaround an issue where `dotnet restore` can [sometimes fail](https://github.com/G-Research/consuldotnet/runs/2204304465).~~

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -3,7 +3,6 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
-    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
-    <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>


marcin-krystianc:
@jgiannuzzi 

Can you split these changes into two separate PRs ? I totally I totally agree with the removal of `NuGet.config`  but I would like to avoid any config manipulation by the CI if it is not absolutely necessary.

.github/workflows/ci.yml
```diff
@@ -45,6 +45,12 @@ jobs:
         uses: actions/setup-dotnet@v1
         with:
           dotnet-version: 3.1.404
+      # WORKAROUND: Reset NuGet sources to default to avoid restore issues
```
marcin-krystianc, line 48: It seems that `setup-dotnet` is somehow broken  https://github.com/actions/setup-dotnet/issues/155). 
The workaround that I would suggest we choose is to add an extra step to clear the http-cache. I've tried this and it succeeded 5 times in a row:
```
# Workaround for https://github.com/actions/setup-dotnet/issues/155
- name: Consul | AspNetCore | Clear http-cache
  run: dotnet nuget list source && dotnet nuget locals http-cache --clear
```

marcin-krystianc:
OK, as per my comment https://github.com/NuGet/Home/issues/10586#issuecomment-809998961, I think that the problem with "Microsoft Visual Studio Offline Packages" is real. 

In my opinion we should avoid special cases in the CI, so the best way to workaround the issue is to not change CI files and keep the `NuGet.config` but just remove the `myget` entry.

jgiannuzzi:
> In my opinion we should avoid special cases in the CI, so the best way to workaround the issue is to not change CI files and keep the `NuGet.config` but just remove the `myget` entry.

Agreed. Amended the PR to do just that.

marcin-krystianc (approved):
👍

PR #76, opened 2021-04-16
Pass WriteOptions parameter to correct parameter of Put(...)

BnMcG:
This was causing token destruction to fail when a per-request ACL token is used.


lorinck (approved):

marcin-krystianc:
Good catch, thanks for fixing this !
I've opened another PR (https://github.com/G-Research/consuldotnet/pull/77) to fix all the remaining call sites and make the internal API less error prone.

PR #80, opened 2021-05-14
CI: Add build matrix

mfkl:
- Refactor configuration properties that could be shared with Directory.Build.Props.
- Enforce a .NET SDK version with global.json for developers.
- Introduce build matrix to test on various .NET TFMs, runtimes and hosts (credits to @jgiannuzzi).

The publish step is the only one which I haven't tested, admittedly.

Just noticed that we have unit tests for Aspnet Core, I will add them to CI runs in a follow up PR (this one may already be large enough).

I've had intermittent test failures, but now I cannot reproduce... Will trigger several GitHub Action runs to try to repro again (never managed to repro locally).

PR #88, opened 2021-05-24
Introduce BaseFixture to try to fix intermittent test failures

mfkl:
sometimes Consul does not initialize fast enough when running parallel tests on the CI. This not elegant workaround aims to address that.

This PR needs to be merged before https://github.com/G-Research/consuldotnet/pull/80

### DIFF
--- a/Consul.Test/BaseFixture.cs
+++ b/Consul.Test/BaseFixture.cs
@@ -49,7 +49,7 @@ namespace Consul.Test
                             _ready = true;
                         }
                     }
-                    catch(OperationCanceledException)
+                    catch (OperationCanceledException)
                     {
                         break;
                     }
--- a/Consul.Test/BaseFixture.cs
+++ b/Consul.Test/BaseFixture.cs
@@ -1,0 +1,50 @@
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Consul.Test
+{
+    public class BaseFixture : IAsyncLifetime
+    {
+        protected ConsulClient _client;
+
+        public BaseFixture()
+        {
+            _client = new ConsulClient(c =>
+            {
+                c.Token = TestHelper.MasterToken;
+                c.Address = TestHelper.HttpUri;
+            });
+        }
+
+        public Task DisposeAsync()
+        {
+            _client.Dispose();
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// With parallel testing on the CI, sometimes the tests start before the consul server finished properly initializing
+        /// This aims to workaround it in a not so elegant way. https://github.com/hashicorp/consul/issues/819
+        /// </summary>
+        /// <returns></returns>
+        public async Task InitializeAsync()
+        {
+            bool ready = false;
+
+            while (!ready)
+            {
+                try
+                {
+                    var leader = await _client.Status.Leader();
+                    if (!string.IsNullOrEmpty(leader))
+                    {
+                        ready = true;
+                    }
+                }
+                catch
+                {
+                }
+            }
+        }
+    }
+}
--- a/Consul.Test/BaseFixture.cs
+++ b/Consul.Test/BaseFixture.cs
@@ -1,3 +1,5 @@
+using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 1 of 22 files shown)


Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,50 @@
...
+            _client.Dispose();
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// With parallel testing on the CI, sometimes the tests start before the consul server finished properly initializing
+        /// This aims to workaround it in a not so elegant way. https://github.com/hashicorp/consul/issues/819
+        /// </summary>
+        /// <returns></returns>
+        public async Task InitializeAsync()
```
marcin-krystianc, line 30: I think we could use some static variable to make the initialisation happen only once for all test classes.

mfkl: Done

Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            bool ready = false;
+
+            while (!ready)
```
marcin-krystianc: I think we need a timeout, otherwise it will wait forever for the consul agent when it is not started.

mfkl: The risk of consul server not starting is rather low I'd say. On the rare chance that it may occur, the CI build will just fail after the GitHub Action times out so not a problem IMO. Let me know.

marcin-krystianc: I'm more concerned about local testing, devs can get very confused when they forget to start Consul before running tests.

mfkl: Done

Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,65 @@
...
+                    try
+                    {
+                        cancelToken.ThrowIfCancellationRequested();
+                        var leader = await _client.Status.Leader();
+                        if (!string.IsNullOrEmpty(leader))
+                        {
+                            _ready = true;
+                        }
+                    }
+                    catch (OperationCanceledException)
```
marcin-krystianc, line 52: If you move `ThrowIfCancellationRequested` outside the `try/catch` block then the `catch (OperationCanceledException)` will not be needed.

Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,65 @@
...
+            _client.Dispose();
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
+        /// With parallel testing on the CI, sometimes the tests start before the consul server finished properly initializing
+        /// This aims to workaround it in a not so elegant way. https://github.com/hashicorp/consul/issues/819
+        /// </summary>
+        /// <returns></returns>
+        public Task InitializeAsync()
```
marcin-krystianc, line 33: I think it is better to avoid non-async methods returning `Task` as they are more difficult to reason about:
```suggestion
        public async Task InitializeAsync()
```

Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            if (_ready) return Task.CompletedTask;
```
marcin-krystianc, line 35:
```suggestion
            if (_ready) return;
```

Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            var cancelToken = new CancellationTokenSource(TimeSpan.FromSeconds(5)).Token;
+
+            var initialize = Task.Run(async () =>
```
marcin-krystianc, line 39:
```suggestion
            await Task.Run(async () =>
```

Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,65 @@
...
+                    {
+                        break;
+                    }
+                    catch
+                    {
+                    }
+                }
+            }, cancelToken);
+
+            return initialize;
```
marcin-krystianc, line 62:
```suggestion

```

marcin-krystianc (approved):
Looks good.

See my suggestions, I think they will simplify the code a little bit more.

marcin-krystianc (approved):

Consul.Test/BaseFixture.cs
```diff
@@ -0,0 +1,59 @@
...
+                        {
+                            _ready = true;
+                        }
+                    }
+                    catch
+                    {
+                    }
+                }
+            }, cancelToken);
+        }
```
jgiannuzzi, line 57: Should the static var `_ready` be protected by a lock as it can be accessed by multiple tests in parallel?

PR #89, opened 2021-05-28
Add CI matrix

mfkl:


.github/workflows/ci.yml
```diff
@@ -5,21 +5,51 @@ on:
...
 
+  # Setup our build matrix
+  setup-build-matrix:
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
+    runs-on: ubuntu-18.04
+    outputs:
+      matrix: ${{ steps.build-matrix.outputs.matrix }}
+    steps:
+      - id: build-matrix
+        uses: jgiannuzzi/setup-build-matrix@v1
```
mfkl, line 21: @jgiannuzzi Can you recall why we need a custom action for setting up the build matrix here?

jgiannuzzi: The idea was to be able to add some combinations to the build matrix that are very difficult to express with a regular GHA matrix (they require a lot of repetition and this custom action makes it easier to manage and maintain).

mfkl: Hmm.. so does this mean we have to maintain https://github.com/jgiannuzzi/setup-build-matrix? It'd be best if we could use the default provided tooling if possible IMO. Less code to maintain the better.

Do you have an old branch with an attempt at building our matrix with the provided GitHub actions? I'd be curious to see the diff.

jgiannuzzi: Here is a before and after dating back from my exploration phase.

### regular GHA matrix

https://github.com/jgiannuzzi/consuldotnet/blob/1e9dcdf0713d3a21ac322ace0b27579dd1ffbcc6/.github/workflows/ci.yml#L10-L40
```yaml
      matrix:
        os: [ubuntu-18.04, windows-2019, macos-10.15]
        component: [Consul, Consul.AspNetCore]
        consul: [1.6.10, 1.7.10, 1.8.6, 1.9.0]
        dotnet: [2.1.811, 3.1.404, 5.0.101]
        exclude:
          - component: Consul.AspNetCore
            consul: 1.7.10
          - component: Consul.AspNetCore
            consul: 1.8.6
          - component: Consul.AspNetCore
            consul: 1.9.0
          - component: Consul.AspNetCore
            dotnet: 2.1.811
        include:
          - os: windows-2019
            component: Consul
            dotnet: "4.7"
            consul: 1.6.10
          - os: windows-2019
            component: Consul
            dotnet: "4.7"
            consul: 1.7.10
          - os: windows-2019
            component: Consul
            dotnet: "4.7"
            consul: 1.8.6
          - os: windows-2019
            component: Consul
            dotnet: "4.7"
            consul: 1.9.0
```
Job output: https://github.com/jgiannuzzi/consuldotnet/actions/runs/412890658
image

### custom-built matrix

https://github.com/jgiannuzzi/consuldotnet/blob/2d71def285a887095004d4f7b632e7940a741b6f/.github/workflows/ci.yml#L22-L42
```yaml
            matrix:
              os: [ubuntu-18.04, windows-2019, macos-10.15]
              component: [Consul, Consul.AspNetCore]
              dotnet: [3.1.404, 5.0.101]
            operations:
              - type: merge
                match: [component]
                matrix:
                  component: [Consul]
                  consul: [1.6.10, 1.7.10, 1.8.6, 1.9.0]
              - type: add
                match: [component]
                matrix:
                  component: [Consul]
                  dotnet: [2.1.811]
              - type: add
                match: [os, component]
                matrix:
                  os: [windows-2019]
                  component: [Consul]
                  dotnet: ["4.7"]
```
Job output: https://github.com/jgiannuzzi/consuldotnet/actions/runs/415967539
image

As you can see, the intent is clearer with the custom-built matrix:
- run on all OSes, for both libraries, with .NET Core 3.1 and .NET 5.0
- for the `Consul` library, test against versions 1.6 through 1.9 of the agent
- for the `Consul` library, test with .NET Core 2.1
- for the `Consul` library on Windows, test with .NET Framework 4.7

With the regular GHA matrix, multiple `exclude` and `include` tricks are used to achieve _almost_ the same result (there is a fake `consul` item with value `1.6.10` that gets created for the `Consul.AspNetCore` build). It also requires maintaining the Consul agent and .NET Framework versions in multiple places.

I delicately weighed the pros and cons of using just the default tooling and avoiding anything custom, but figured the pros outweighed the cons. Ideally this functionality would get directly built into GHA in the future, but it has to start somewhere 😄

mfkl: Thank you for the detailed explanation.

I agree that the final YML configuration is more readable and explicit with your custom action. 

However, I don't know the implications of maintaining https://github.com/jgiannuzzi/setup-build-matrix. 
For example, how often does GitHub runner updates would break our custom action? How much work will it be to fix it? What about versioning? Dependencies? Security? etc. 

[That's](https://github.com/jgiannuzzi/setup-build-matrix/blob/main/dist/index.js) 15 thousands lines of JS I'm personally unfamiliar with :-) Did you get that initial JS from somewhere? I couldn't find the default GitHub Action Matrix code anywhere (tried [this](https://github.com/marketplace?type=actions&query=matrix+), and [that](https://github.com/sdras/awesome-actions)).

Personally, I would probably stick to the default provided tooling even if the resulting config is less readable. Just my 2 cents :-)

.github/workflows/ci.yml
```diff
@@ -53,69 +99,64 @@ jobs:
...
         run: |
           cd Consul.Test
           SYSTEM=$(uname | sed 's/MINGW.*/windows/' | tr A-Z a-z)
-          curl -sSL https://releases.hashicorp.com/consul/${CONSUL_VERSION}/consul_${CONSUL_VERSION}_${SYSTEM}_amd64.zip -o consul.zip
+          curl -sSL https://releases.hashicorp.com/consul/${{ matrix.consul }}/consul_${{ matrix.consul }}_${SYSTEM}_amd64.zip -o consul.zip
           unzip consul.zip
           rm consul.zip
-      - name: Build
-        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true
+      - name: Build library
```
marcin-krystianc, line 110: There is no need to build library explicitly, it is enough to build tests (required libraries will be built automatically). If we remove this step we can also remove `libframework`.

.github/workflows/ci.yml
```diff
@@ -53,69 +99,64 @@ jobs:
...
         run: |
-          cd Consul.Test
-          ./consul agent -dev -config-file test_config.json -log-file consul.log >consul-stdout.log 2>consul-stderr.log &
-          dotnet test --configuration=Release --logger "GitHubActions;report-warnings=false" --no-build -v=Normal
+          cd ${{ matrix.component }}.Test
+          if [ "${{ matrix.component }}" == "Consul" ]
+          then
+            ./consul agent -dev -config-file test_config.json -log-file consul.log >consul-stdout.log 2>consul-stderr.log &
+          fi
+          if [ "${{ steps.dotnet.outputs.framework }}" != "netcoreapp2.1" ]
```
marcin-krystianc, line 122: This check is not needed, `GitHubActions` is supported on all frameworks.

.github/workflows/ci.yml
```diff
@@ -53,69 +99,64 @@ jobs:
...
-            $suffix = 'g' + $(git rev-parse --short HEAD)
-            $params = "--version-suffix", $suffix
-          }
-
-          dotnet pack --configuration=Release --no-build @params Consul/Consul.csproj
-
-          if ("${{ github.ref }}" -like "refs/tags/v*") {
-            $tag = "${{ github.ref }}".SubString(11)
-            $expectedFile = "dist/consul/Consul.$tag.nupkg"
-            # Check whether the tag and the package version match together
```
marcin-krystianc, line 94: Previously we were checking whether the tag version and the package version match together, but now this functionality is missing.

.github/workflows/ci.yml
```diff
@@ -53,69 +99,64 @@ jobs:
...
+            logger='--logger "GitHubActions;report-warnings=false"'
+          fi
+          dotnet test --framework ${{ steps.dotnet.outputs.framework }} --configuration=Release --logger "GitHubActions;report-warnings=false" --no-build -v=Normal
       - name: Upload Consul logs
-        if: failure()
+        if: failure() && matrix.component == 'Consul'
         uses: actions/upload-artifact@v2
         with:
-          name: consul-logs-${{ matrix.os }}
+          name: consul-logs-${{ matrix.os }}_dotnet-${{ steps.dotnet.outputs.version }}_consul-${{ matrix.consul }}
```
marcin-krystianc, line 131: Why we use `dotnet.outputs.version` here, isn't `dotnet.outputs.framework` a better choice ?

.github/workflows/ci.yml
```diff
@@ -5,21 +5,51 @@ on:
...
+                match: [component]
+                matrix:
+                  component: [Consul]
+                  dotnet: ['netcoreapp2.1:5.0.202:experimental']
+              - type: add
+                match: [os, component]
+                matrix:
+                  os: [windows-2019]
+                  component: [Consul]
+                  dotnet: ['net461:4.6.1']
```
marcin-krystianc, line 44: We now support net461 on all systems.

.github/workflows/ci.yml
mfkl: Good catch

.github/workflows/ci.yml
jgiannuzzi: The actual code is in [`lib/main.js`](https://github.com/jgiannuzzi/setup-build-matrix/blob/main/lib/main.js) and is 97 lines of original hand-crafted JS 😅
They are themselves a rewrite of 46 lines of [original hand-crafted Python](https://github.com/jgiannuzzi/setup-build-matrix/blob/ecdb44368d552cf5ed422edd0d06e87d06e1b9ab/setup-build-matrix.py).

The project relies on a few JS dependencies and I should totally use dependabot to keep them up-to-date. In terms of breaking changes from GitHub Actions, the features this action relies on are probably going to be supported 'forever', as they are fundamental to most workflows. It uses the first-party language for creating actions (JS) and builds a string output from a string input. That string output is a JSON-serialised matrix, flattened to be a list KV pairs passed as `include`. GitHub would break a lot of existing workflows and actions if they changed any of that.

The GitHub Action matrix code is unfortunately not open source. [Here](https://github.com/nektos/act/blob/33ccfa6f3b85bfe02716e3369f19246de450c9e7/pkg/model/workflow.go#L196) is a compatible implementation for reference.

As the author of `setup-build-matrix`, I am of course biased, so I will let @marcin-krystianc and @mfkl decide what would be best for Consul.NET.

mfkl:
I'm still seeing intermittent test failures

```
[xUnit.net 00:00:10.54]     Consul.Test.KVTest.KV_WatchGet_Cancel [FAIL]
[xUnit.net 00:00:10.54]       A cancellation exception was not thrown when one was expected.
[xUnit.net 00:00:10.54]       Expected: True
[xUnit.net 00:00:10.54]       Actual:   False
[xUnit.net 00:00:10.54]       Stack Trace:
[xUnit.net 00:00:10.54]         /home/runner/work/consuldotnet/consuldotnet/Consul.Test/KVTest.cs(269,0): at Consul.Test.KVTest.KV_WatchGet_Cancel()
[xUnit.net 00:00:10.54]         --- End of stack trace from previous location ---
```

.github/workflows/ci.yml
mfkl: Just updated the branch with Marcin's solution that uses default tooling so we can compare (previous code with your custom action [here](https://github.com/mfkl/consuldotnet/blob/mega-ci5/.github/workflows/ci.yml). 

Personally I favor using the default tooling approach.

.github/workflows/ci.yml
```diff
@@ -57,65 +60,67 @@ jobs:
...
     steps:
       - name: Download NuGet package artifact
         uses: actions/download-artifact@v2
         with:
-          name: nuget-package
-          path: dist/consul
+          path: dist
       - name: Publish to NuGet
-        run: dotnet nuget push dist/consul/*.nupkg --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
+        run: dotnet nuget push dist/*.nupkg --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
```
marcin-krystianc, line 126: I think we cannot push `Consul.AspNetCore` package yet, So maybe let's use more specific path here:
```suggestion
        run: dotnet nuget push dist/Consul.1*.nupkg --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
```

.github/workflows/ci.yml
```diff
@@ -5,46 +5,49 @@ on:
...
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Setup .NET Core SDK v2.1.x
+        if: matrix.framework == 'netcoreapp2.1'
         uses: actions/setup-dotnet@v1
         with:
           dotnet-version: 2.1.x
```
marcin-krystianc, line 48: I'm not sure whether we should be using pinned or floating version here.
With pinned versions our tests are repeatable, but with floating versions we get updates to latest patch version for free.
I think either is fine.

marcin-krystianc (commented):
I think it looks sensible 👍

.github/workflows/ci.yml
mfkl: Ok

.github/workflows/ci.yml
```diff
@@ -57,65 +60,67 @@ jobs:
...
     steps:
       - name: Download NuGet package artifact
         uses: actions/download-artifact@v2
         with:
-          name: nuget-package
-          path: dist/consul
+          path: dist
       - name: Publish to NuGet
-        run: dotnet nuget push dist/consul/*.nupkg --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
+        run: dotnet nuget push dist/Consul.nupkg --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
```
marcin-krystianc, line 126: Package file name contains also a version, e.g.: `consul.1.6.1.1.nupkg`

mfkl: btw `"${{ github.ref }}".SubString(11)` this doesn't look so future-proof hmm

marcin-krystianc (dismissed):

.github/workflows/ci.yml
marcin-krystianc: It makes use of a fact that the git ref starts with `refs/tags/v`. It is not pretty, but I think it is ok

mfkl: Ok

mfkl:
One more thing I need to do is to edit the `KV_WatchGet_Cancel` test to increase timing (because it will intermittently fail otherwise).

mfkl:
The 2 sometimes failing tests I'm looking into

```
Error: Consul.Test.SemaphoreTest.Semaphore_OneShot: System.InvalidOperationException : Failed to create session
  Failed Consul.Test.SemaphoreTest.Semaphore_OneShot [365 ms]
  Error Message:
   System.InvalidOperationException : Failed to create session
---- Consul.ConsulRequestException : Unexpected response, status code InternalServerError: Missing node registration
  Stack Trace:
     at Consul.Semaphore.Acquire(CancellationToken ct) in D:\a\consuldotnet\consuldotnet\Consul\Semaphore.cs:line 343
   at Consul.Semaphore.Acquire(CancellationToken ct) in D:\a\consuldotnet\consuldotnet\Consul\Semaphore.cs:line 456
   at Consul.Test.SemaphoreTest.Semaphore_OneShot() in D:\a\consuldotnet\consuldotnet\Consul.Test\SemaphoreTest.cs:line 96
--- End of stack trace from previous location where exception was thrown ---
----- Inner Stack Trace -----
   at Consul.PutRequest`2.Execute(CancellationToken ct) in D:\a\consuldotnet\consuldotnet\Consul\Client_PutRequests.cs:line 352
   at Consul.Session.Create(SessionEntry se, WriteOptions q, CancellationToken ct) in D:\a\consuldotnet\consuldotnet\Consul\Session.cs:line 296
   at Consul.Semaphore.CreateSession() in D:\a\consuldotnet\consuldotnet\Consul\Semaphore.cs:line 668
   at Consul.Semaphore.Acquire(CancellationToken ct) in D:\a\consuldotnet\consuldotnet\Consul\Semaphore.cs:line 337


   [xUnit.net 00:00:10.76]     Consul.Test.KVTest.KV_WatchList_Cancel [FAIL]
[xUnit.net 00:00:10.76]       A cancellation exception was not thrown when one was expected.
[xUnit.net 00:00:10.76]       Expected: True
[xUnit.net 00:00:10.76]       Actual:   False
[xUnit.net 00:00:10.76]       Stack Trace:
[xUnit.net 00:00:10.76]         D:\a\consuldotnet\consuldotnet\Consul.Test\KVTest.cs(335,0): at Consul.Test.KVTest.KV_WatchList_Cancel()
[xUnit.net 00:00:10.76]         --- End of stack trace from previous location where exception was thrown ---
```

mfkl:
After several runs, it appears that the tests no longer fail with these 2 last commits.

marcin-krystianc (dismissed):

.github/workflows/ci.yml
```diff
@@ -57,65 +60,69 @@ jobs:
...
+          # WORKAROUND: Add letter prefix to ensure that MSBuild treats
+          # the suffix as a string. e.g. '0313071' will fail as an invalid
+          # version string, but 'G0313071' will succeeded. It looks like
+          # the parser does not like numbers with a leading zero.
+          $suffix = 'g' + $(git rev-parse --short HEAD)
+          $params = "--version-suffix", $suffix
+        }
 
-          if ("${{ github.ref }}" -like "refs/tags/v*") {
+        dotnet pack --configuration=Release --output dist
```
jgiannuzzi, line 97: I think we're missing `@params` here

.github/workflows/ci.yml
```diff
@@ -5,46 +5,49 @@ on:
...
-    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        framework: [netcoreapp3.1]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
```
jgiannuzzi, line 22: I think we should add a step to install the latest .NET 5.0.x SDK to have consistency between operating systems, otherwise we will be at the mercy of the runners being upgraded at their own pace.

.github/workflows/ci.yml
```diff
@@ -5,46 +5,49 @@ on:
...
-      matrix:
+      matrix:      
+        consul: [1.6.10, 1.7.10, 1.8.6, 1.9.0]
+        framework: [net461, netcoreapp2.1, netcoreapp3.1, net5.0]
         os: [ubuntu-18.04, windows-2019, macos-10.15]
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
```
jgiannuzzi, line 43: I think we should add a step to install the latest .NET 5.0.x SDK to have consistency between operating systems, otherwise we will be at the mercy of the runners being upgraded at their own pace.

.github/workflows/ci.yml
```diff
@@ -57,65 +60,69 @@ jobs:
...
-            # version string, but 'G0313071' will succeeded. It looks like
-            # the parser does not like numbers with a leading zero.
-            $suffix = 'g' + $(git rev-parse --short HEAD)
-            $params = "--version-suffix", $suffix
-          }
+          name: consul-logs-${{ matrix.consul }}-${{ matrix.framework }}-${{ matrix.os }}
+          path: consul*.log
 
-          dotnet pack --configuration=Release --no-build @params Consul/Consul.csproj
+  package:
```
jgiannuzzi, line 80: `package` and `publish` won't work once we want to release Consul.AspNetCore, which is a regression from the previous CI. I know we can't yet publish Consul.AspNetCore (I need to talk to the current package owner), but we should keep our CI ready for it.

jgiannuzzi (changes requested):
Hey @mfkl, this is awesome!

I just have a couple of comments below that I think we should address before merging.

.github/workflows/ci.yml
jgiannuzzi: Thanks to @zlzforever, we now have ownership of Consul.AspNetCore on nuget.org, so this really needs to work now 😄

.github/workflows/ci.yml
mfkl: Fixed

.github/workflows/ci.yml
mfkl: Hmm we don't test net5.0 for consul so that is why it is not installed. Added it.

.github/workflows/ci.yml
mfkl: It is already https://github.com/G-Research/consuldotnet/blob/c1bb616cac7c508ee878f8aeab2ed1eb52040bbd/.github/workflows/ci.yml#L59

.github/workflows/ci.yml
mfkl: Not really a regression technically, as there was no continuous delivery for the asp integration :-P

Anyway, I sort of fixed that now. But we're going to need to decide on the deploy process here. If one tag and shared version number for both packages, it is easy. If packages have different package versions, different automated releases, and/or different tags, well... some decisions need to be made :-)

.github/workflows/ci.yml
jgiannuzzi: sorry, my bad!

.github/workflows/ci.yml
jgiannuzzi: Because of `global.json`, we were building it with the .NET 5 SDK, hence my comment.
But it's even better that we test it with .NET 5 now!

.github/workflows/ci.yml
jgiannuzzi: There is/was a CD for it, we just never pulled the trigger on making a release 😔
https://github.com/G-Research/consuldotnet/blob/2e7b1b15c8bac03d1ff1b8f47ef578567ffea3f3/.github/workflows/consul.aspnetcore.yml#L62-L76

The decision back then was to use a different tag for releasing the ASP library (`Consul.AspNetCore_vX.Y.Z`), but I'm more than open to revisit this.

The way is see it at the moment is that using the same version numbers would probably mean having many releases of the ASP library that have no actual changes. On the other hand, having different version numbers means using those ugly tags.

If we do go for different version numbers, I think this 'initial' release should be 2.0.0, as the existing 1.0.0 package was a different project altogether.

@marcin-krystianc @mfkl thoughts?

.github/workflows/ci.yml
jgiannuzzi: I think if we add scheduled runs of the tests on `master`, we would get quickly alerted of any problems with newer versions. The most important is that we should now have consistency of .NET versions between the various OS (as opposed to a few months ago, when a partial rollout of the latest .NET 5 SDK to the macOS runners got us into trouble).

.github/workflows/ci.yml
mfkl: > The way is see it at the moment is that using the same version numbers would probably mean having many releases of the ASP library that have no actual changes.

True, but that also means they are kept in sync by design, and provides a straightforward versioning scheme to end user.

My opinion is to have them share the version numbers, if I recall correctly Marcin prefers the different version number approach. Anyway, it is not highly critical, I guess :-)

marcin-krystianc: Both approaches have pros & cons. I think we are at the point were we can try using the same version for both libraries (1.x) and if we decide later that we don't like it we can change the version of AspNetCore to v2.x. But the other order is not possible, once we publish version 2.x we will not be able to switch back to v1.x...
Thus maybe we should try using the same version for both libraries and see how it goes ?

jgiannuzzi: Sounds sensible to me, let's go for same version!

.github/workflows/ci.yml
```diff
@@ -57,65 +65,70 @@ jobs:
...
 
   # Publish NuGet package when a tag is pushed.
-  # Tests need to succeed on all platforms first, including having a tag name that matches the version number.
+  # Tests need to succeed for all components and on all platforms first,
+  # including having a tag name that matches the version number.
   publish:
-    if: startsWith(github.ref, 'refs/tags/v')
-    needs: build
-    runs-on: ubuntu-18.04
+    if: github.ref == 'refs/heads/master' && startsWith(github.ref, 'refs/tags/v')
```
jgiannuzzi, line 122: I don't think this will ever get triggered - any reason you changed this condition?

.github/workflows/ci.yml
```diff
@@ -57,65 +65,70 @@ jobs:
...
+          $params = "--version-suffix", $suffix
+        }
+
+        dotnet pack --configuration=Release --output dist @params
 
-          if ("${{ github.ref }}" -like "refs/tags/v*") {
+        if ("${{ github.ref }}" -like "refs/tags/v*") {
             $tag = "${{ github.ref }}".SubString(11)
-            $expectedFile = "dist/consul/Consul.$tag.nupkg"
+            $expectedFile = "dist/Consul.$tag.nupkg"
```
jgiannuzzi, line 106: We should also check that the Consul.AspNetCore package has the right name.

.github/workflows/ci.yml
```diff
@@ -57,65 +65,70 @@ jobs:
...
             # Check whether the tag and the package version match together
             if (-not (Test-Path -Path $expectedFile)) {
                 echo "Expected file $expectedFile doesn't exist"
```
jgiannuzzi, line 109: We should output a proper workflow error here (see the [docs](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message))

```suggestion
                echo "Expected file $expectedFile doesn't exist"
                echo "::error ::Expected file $expectedFile doesn't exist"
```

.github/workflows/ci.yml
mfkl: Ok

.github/workflows/ci.yml
mfkl: > I don't think this will ever get triggered

Why not? It only checks that we are on master.

> any reason you changed this condition from the previous?

Just an extra defensive check to avoid this: https://github.com/mfkl/consuldotnet/actions/runs/967076197

If someone or something creates pushes a new branch to this repo, and tags it, the CI would try to deploy to NuGet. I can remove it, if you prefer.

.github/workflows/ci.yml
mfkl: Done

.github/workflows/ci.yml
mfkl: Done

.github/workflows/ci.yml
jgiannuzzi: I think this won't work because `github.ref` cannot both be a branch and a tag. Pushing to a branch will trigger a workflow, and pushing to a tag will trigger a different one.

In my opinion, we should just go for the same check as before.

Then in the future we should start using GitHub Actions Environments and the `workflow_run` event to ensure only a tag on `master` can push to a package to NuGet. This is something that our DevOps team could add to this repo.

mfkl: Fair enough.

jgiannuzzi (approved):
Awesome work, thanks @mfkl!

.github/workflows/ci.yml
```diff
@@ -5,46 +5,54 @@ on:
...
 env:
-  CONSUL_VERSION: 1.6.10
-
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+  DOTNET_CLI_TELEMETRY_OPTOUT: true
+    
 jobs:
-
-  # Enforces the consistency of code formatting using `.editorconfig` and the `dotnet-format` tool.
-  check-format:
```
jgiannuzzi, line 13: Sorry for the last minute change request @mkfl, but I just noticed that this PR is removing the `check-format` job!
Could you please add it back?

jgiannuzzi (changes requested):

marcin-krystianc:
Looks good apart from the missing check-format job,

.github/workflows/ci.yml
mfkl: Done

jgiannuzzi (approved):

marcin-krystianc (approved):

PR #107, opened 2021-07-16
Enable codeql analysis

ljubon:
Example run: https://github.com/ljubon/consuldotnet/actions/runs/1036400778
View alerts: https://imgur.com/a/3KT6XWD


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,98 @@
...
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'csharp' ]
+        framework: [netcoreapp3.1, net5.0]
+        consul: [1.6.10, 1.7.14, 1.8.13, 1.9.7, 1.10.0]
```
jgiannuzzi, line 37: I think there is no need to use a matrix for `framework` and `consul`. `framework` is used in the main CI pipeline to parallelise jobs, and `consul` is used to determine which version of the Consul agent to run the tests against. Here we don't run the tests, but instead we just want to build the entire solution for all frameworks (for which autobuild should actually be enough I guess).

mfkl:
Examples of CodeQL errors on Consul server source code on lgtm.com (powered by CodeQL) https://lgtm.com/projects/g/hashicorp/consul?mode=list

mfkl (dismissed):

jgiannuzzi (changes requested):
I would like the build step to be changed to either be autobuild or a very simple build of the entire solution for all farther frameworks with the latest .NET 5 SDK. No need to download Consul server.

ljubon:
> I would like the build step to be changed to either be autobuild or a very simple build of the entire solution for all farther frameworks with the latest .NET 5 SDK. No need to download Consul server.

hellou, it seems that `auto build works just fine
I've checked, it's using latest.NET 5.0
e.g GH Action https://github.com/ljubon/consuldotnet/runs/3149516224?check_suite_focus=true

jgiannuzzi (approved):

jgiannuzzi:
Thanks for your contribution @ljubon!

PR #110, opened 2021-07-19
enable dependabot

ljubon:
Enable DependaBot security feature
Useful links:
- https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/about-alerts-for-vulnerable-dependencies
- https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/about-managing-vulnerable-dependencies#dependabot-alerts

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  # Enable version updates for nuget
+  - package-ecosystem: "nuget"
+    # Look for NuGet dependency info from the `root` directory
+    directory: "/"
+    # Check the nuget registry for updates every day (weekdays) schedule:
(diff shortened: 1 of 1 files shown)


mfkl:
I do not think dependabot is necessary on consuldotnet:
- Because this is a library, we want to be really careful about upgrading dependencies versions. Potentially the opposite behavior for an app.
- Consuldotnet has very little dependencies, so a bot seems like unnecessary overhead.

marcin-krystianc:
As @mfkl had explained we don't want automatic updates for library projects. We discussed an idea to limit dependabot to run only for test projects but that seems to be not possible. I've tried such config:
```
version: 2
updates:
  - package-ecosystem: "nuget"
    directory: "/Consul.AspNetCore.Test"
    schedule:
      interval: "daily"
  - package-ecosystem: "nuget"
    directory: "/Consul.Test"
    schedule:
      interval: "daily"
```
It unfortunately still created a PR to update the `Newtonsoft.Json` in `Consul.csproj` which is exactly what we wanted to avoid:
image

PR #111, opened 2021-07-28
Added support for TaggedAddresses for services

evgslyusar:
Added support for TaggedAddresses for services, according to the contacts [consul API](https://www.consul.io/api-docs/catalog#servicetaggedaddresses)


mfkl (approved):
LGTM

Consul/Agent.cs
```diff
@@ -145,6 +145,8 @@ public class AgentService
         public string[] Tags { get; set; }
         public int Port { get; set; }
         public string Address { get; set; }
+
+        public IDictionary<string, ServiceTaggedAddress> TaggedAddresses { get; set; }
```
marcin-krystianc, line 149: I know that it is technically a dictionary, but on other hand only two keys (`lan` and `wan`) can be used. I wonder if it wouldn't be better if it is defined as a struct so there is no doubt what can and what cannot be a key?:
```
public TaggedAddresses TaggedAddresses { get; set; }
```

```
    public class TaggedAddress
    {
        public string Address { get; set; }
        public int Port { get; set; }
    }

    public class TaggedAddresses
    {
        public TaggedAddress Lan { get; set; }
        public TaggedAddress Wan { get; set; }
    }
 ```
 
 Initialisation of such struct is a little bit easier:
 ```
 var taggedAddresses = new ServiceTaggedAddresses
 {
      Lan = new ServiceTaggedAddress {Address = "127.0.0.1", Port = 80},
      Wan = new ServiceTaggedAddress {Address = "192.168.10.10", Port = 8000},
 }
 ```

File: Consul.Test/CatalogTest.cs [hunk 2]
```diff
@@ -140,6 +146,58 @@ public async Task Catalog_RegistrationDeregistration()
...
+            await _client.Catalog.Deregister(dereg);
+
+            dereg = new CatalogDeregistration
+            {
+                Datacenter = "dc1",
+                Node = "foobar",
+                Address = "192.168.10.10"
+            };
+
+            await _client.Catalog.Deregister(dereg);
```
marcin-krystianc, line 193: I think that deregistration is not essential for this test. Should we remove it entirely?

mfkl: Noticed that too but it's been done that way for the other tests so this PR does stick to the existing style. Fine either way

mfkl:
Could you please update the changelog file as well?

Consul/Agent.cs
evgslyusar: Yes, WAN and LAN are specified as keys but it is not the restriction, we can use any keys. For example, we can use 'Internal', 'external' keys. Moreover, in a service registration process via a configuration file, there are no restrictions on keys.

the example below works amazing:

```
          var svcID = KVTest.GenerateTestKeyName();
          var registration = new AgentServiceRegistration
            {
                Name = svcID,
                Port = 8000,
                TaggedAddresses = new Dictionary<string, ServiceTaggedAddress>
                {
                    {"lan", new ServiceTaggedAddress {Address = "127.0.0.1", Port = 80}},
                    {"wan", new ServiceTaggedAddress {Address = "192.168.10.10", Port = 8000}},
                    {"internal", new ServiceTaggedAddress {Address = "127.0.0.2", Port = 82}},
                    {"external", new ServiceTaggedAddress {Address = "192.168.10.12", Port = 8002}}
                }
            };

```

marcin-krystianc: All right, so let's keep it as dictionary then 👍

Consul.Test/CatalogTest.cs
marcin-krystianc: I think it is done this way only in the `Catalog_RegistrationDeregistration` which makes sense. I believe it is better to remove it from this newly added tests if it is not necessary.

marcin-krystianc (commented):
Looks really good, I would only remove the extra code from the `Catalog_GetTaggedAddressesService` tests.

Consul.Test/CatalogTest.cs
evgslyusar: done

mfkl (approved):

Consul.Test/CatalogTest.cs
marcin-krystianc: @evgslyusar 
I see that You removed extra code form `Catalog_RegistrationDeregistration` method (https://github.com/G-Research/consuldotnet/pull/111/commits/aafad6f715a6b1f10b376c1e348ac0e3292d233a). I was more talking about lines 171..188 in the `Catalog_GetTaggedAddressesService` method. Can you remvoe them as well ?

mfkl:
Just need to remove the deregistration lines and this should be good to go.

ruxo:
Sorry, any chance this change is going to get merged?

marcin-krystianc:
@evgslyusar,  thank you for your contribution.

PR #127, opened 2021-10-20
Introducing a bug report form

pavlovic-ivan:
Hi team. I've taken the current bug report structure, and created another, with the same fields, but this time, you will see that creating a bug report is actually a proper form with fields (some are required, some optional).

This should give you the ability to enforce users to provide necessary data that you need for the bug report. Also, labels can be preset. Assignees can be preset. And there are features available as well, eg: i can add a field that will format in log format (as i've seen you already ask for logs screenshot).

Let me know if you like the idea, or if there are any suggestions.

Best regards,
Ivan


pavlovic-ivan:
@mfkl Thanks for noticing the whitespace. Please review again. Whitespace is removed from the name, and the `bug_report.md` is now removed.

mfkl (approved):

marcin-krystianc (approved):

mfkl:
Thanks @pavlovic-ivan!

pavlovic-ivan:
No, thank you @mfkl :)

PR #134, opened 2022-02-03
Test with Consul 1.11.x

jgiannuzzi:
Add support for testing against Consul agent 1.11.x.

### DIFF
--- a/Consul.Test/ACLTest.cs
+++ b/Consul.Test/ACLTest.cs
@@ -30,6 +30,9 @@ namespace Consul.Test
         public async Task ACL_CreateDestroyLegacyToken()
         {
             Skip.If(string.IsNullOrEmpty(TestHelper.MasterToken));
+            var info = await _client.Agent.Self();
+            var version = new Version(info.Response["Config"]["Version"]);
+            Skip.If(version >= new Version("1.11.0"), "Legacy ACL system was removed in Consul 1.11.");
 
 #pragma warning disable CS0618 // Type or member is obsolete
             var aclEntry = new ACLEntry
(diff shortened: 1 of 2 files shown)


Consul.Test/ACLTest.cs
```diff
@@ -30,6 +30,9 @@ public class ACLTest : BaseFixture
         public async Task ACL_CreateDestroyLegacyToken()
         {
             Skip.If(string.IsNullOrEmpty(TestHelper.MasterToken));
+            var info = await _client.Agent.Self();
+            var version = new Version(info.Response["Config"]["Version"]);
+            Skip.If(version >= new Version("1.11.0"), "Legacy ACL system was removed in Consul 1.11.");
```
jgiannuzzi, line 35: I have done this in the same way we skip tests in `AuthMethodTest.cs`, where it's repeated in each test:

https://github.com/G-Research/consuldotnet/blob/9f874844b42ced7c971447c0a461abe66e25d911/Consul.Test/AuthMethodTest.cs#L32-L33

I do wonder if this might instead live at the class level, or in `TestHelper.cs`. Thoughts?

marcin-krystianc: I think that the best would be to create new method in ACLTest class to avoid code duplication.

e.g:
```
  public class ACLTest : BaseFixture
  {
        async Task SkipIfAclNotSupportedAsync()
        {
            var info = await _client.Agent.Self();
            var version = new Version(info.Response["Config"]["Version"]);
            Skip.If(version >= new Version("1.11.0"), "Legacy ACL system was removed in Consul 1.11.");
        }

        [SkippableFact]
        public async Task ACL_CreateDestroyLegacyToken()
        {
            Skip.If(string.IsNullOrEmpty(TestHelper.MasterToken));
            await SkipIfAclNotSupportedAsync();
...
```

PS:

jgiannuzzi: Agreed 👍
Implemented in 1cf9460.

marcin-krystianc (approved):
LGTM

PR #135, opened 2022-02-03
Add support for building and testing with .NET 6

jgiannuzzi:
This PR adds support for building and testing with .NET 6.
`global.json` is updated to require .NET 6, and the builtin `dotnet format` is used instead of the external tool.


marcin-krystianc (approved):
nice 👍

PR #136, opened 2022-03-04
CI: Pin .NET SDK 6.0.102 for now to avoid dotnet format regression

mfkl:
Fixes https://github.com/G-Research/consuldotnet/runs/5330947266?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup .NET SDK v6.0.x
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: 6.0.102
       - name: Check format
         run: dotnet format --verify-no-changes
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup .NET SDK v6.0.x
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: 6.0.102
```
marcin-krystianc, line 22: Can you add a comment why we are pinning dotnet version ? You can also include a link to dotnet issue.

mfkl:
It's all in the commit message, but I can add a comment in the code as well
if you prefer.

marcin-krystianc:
> It's all in the commit message, but I can add a comment in the code as well if you prefer.

Yeah, I think it is good to include such information in the code as well. So next person who sees it knows the reason for it without using extra tool like `git blame`.

mfkl:
Ok

marcin-krystianc (approved):

jgiannuzzi (approved):

PR #139, opened 2022-04-21
Auto-update Consul versions

jgiannuzzi:
This PR introduces a new daily scheduled workflow that checks for the latest versions of Consul and updates our CI with them if needed. It will do so by creating a new branch with the changes and opening a PR, which is much better than me doing all of that manually every once in a while 😅

As the default token used by GitHub Actions does not have the special `workflows` permissions required to update our CI, a custom GitHub App had to be created instead with the required permissions. Its credentials have been set as secrets within the `update-consul-versions` environment that only `master` can access.

An example of a PR created by this workflow can be found here: https://github.com/jgiannuzzi/consuldotnet/pull/2

### DIFF
--- a/.github/workflows/scripts/update-consul-versions.ps1
+++ b/.github/workflows/scripts/update-consul-versions.ps1
@@ -1,0 +1,32 @@
+$workflow = ".github/workflows/ci.yml"
+$owner = "hashicorp"
+$repo = "consul"
+
+# Suppress telemetry reminder
+Set-GitHubConfiguration -SuppressTelemetryReminder
+
+# Use GITHUB_TOKEN if provided
+if ($Env:GITHUB_TOKEN -ne $null)
+{
+  $token = ($Env:GITHUB_TOKEN | ConvertTo-SecureString -AsPlainText -Force)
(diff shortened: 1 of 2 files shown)


mfkl:
How do we decide the range of older versions we keep supporting? At which point will we say "hmm this is too old, for reason X/Y we can't support it"?

And what happens regarding this script at this point?

Really cool to not have to do so manually anyway :)

jgiannuzzi:
> How do we decide the range of older versions we keep supporting? At which point will we say "hmm this is too old, for reason X/Y we can't support it"?

Hashicorp seems to maintain the last 3 minor releases, so arguably that is what we should support. Maybe we could be a bit more conservative and support the last 4 or 5.
_However_, until we close the gap in terms of API support and finally update our minor version, it would be fair for our users to expect our current 1.6 release to support at least Consul 1.6. So until then, I would say we should support 1.6+ (so the last 7 releases currently — 1.6 -> 1.12).
Once we decide to drop some releases, the script should be trivial to adapt (I even wrote it in PowerShell to cater to this project's audience 😅).

marcin-krystianc (approved):
Nice !

PR #146, opened 2022-05-19
Fix deadlock in .net framework.

bosima:
Deadlock occurs when calling ‘Services’ method synchronously in .Net Framework, which is a classic problem.

None of the other methods in this class use await, so I removed await.

### DIFF
--- a/Consul/Agent.cs
+++ b/Consul/Agent.cs
@@ -374,9 +374,9 @@ namespace Consul
         /// Services returns the locally registered services
         /// </summary>
         /// <returns>A map of the registered services and service data</returns>
-        public async Task<QueryResult<Dictionary<string, AgentService>>> Services(CancellationToken ct = default(CancellationToken))
+        public Task<QueryResult<Dictionary<string, AgentService>>> Services(CancellationToken ct = default(CancellationToken))
         {
-            return await Services(null, ct);
+            return Services(null, ct);
         }
 
         /// <summary>
@@ -384,9 +384,9 @@ namespace Consul
         /// </summary>
         /// <param name="filter">Specifies the expression used to filter the queries results prior to returning the data</param>
         /// <returns>A map of the registered services and service data</returns>
-        public async Task<QueryResult<Dictionary<string, AgentService>>> Services(Filter filter, CancellationToken ct = default(CancellationToken))
+        public Task<QueryResult<Dictionary<string, AgentService>>> Services(Filter filter, CancellationToken ct = default(CancellationToken))
         {
-            return await _client.Get<Dictionary<string, AgentService>>("/v1/agent/services", null, filter).Execute(ct);
+            return _client.Get<Dictionary<string, AgentService>>("/v1/agent/services", null, filter).Execute(ct);
         }
 
         /// <summary>


Consul/Agent.cs
```diff
@@ -374,19 +374,19 @@ public async Task<string> GetNodeName(CancellationToken ct = default(Cancellatio
         /// Services returns the locally registered services
         /// </summary>
         /// <returns>A map of the registered services and service data</returns>
-        public async Task<QueryResult<Dictionary<string, AgentService>>> Services(CancellationToken ct = default(CancellationToken))
+        public Task<QueryResult<Dictionary<string, AgentService>>> Services(CancellationToken ct = default(CancellationToken))
         {
-            return await Services(null, ct);
+            return Services(null, ct);
```
marcin-krystianc, line 379: Removing async / await is not going to change the fact that this method is asynchronous (it still returns an asynchronous task). Any problem you have with deadlocks will remain the same. 
The downside of removing async / await from a method is that in case of exception being thrown from the `Services` method, method `Services(CancellationToken ct = default(CancellationToken))` will not be on the call stack.
There is a nice blog post about all pitfalls when async / await is removed from a method: https://blog.stephencleary.com/2016/12/eliding-async-await.html.

marcin-krystianc (commented):
@bosima As far as I know it is not possible to call asynchronous methods synchronously and avoid deadlocks. It is not only .NET Framework problem but it is the same for .NET Core too.

Your change doesn't help.  Although you removed async / await keywords, `Services` method is still asynchronous (it returns a task which needs to be awaited at some point).

Consul/Agent.cs
bosima: Sorry, I skipped a lot of details.

I understand that if async is used, it should be all async. This is best practice. But we have some old asp.net programs, they use synchronous method. We used it like this before:

    var services = Services.Result;

It's ok before v1.6.10.4, deadlock occurred in V1 6.10.4 and later.

The reason of the deadlock is the contention of the synchronization context，but _AspNetSynchronizationContext_ was removed in ASP.NET Core. reference: 
https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html

If  want to use _await_,  in order to be compatible with .net framework, the safer way is to add _ConfigureAwait(false)_:

    return await Services(null, ct).ConfigureAwait(false);

But other methods is ok that not use _await_ in _Agent_,  because they just return _Task_, not execute.  Asynchronous calls in _Execute_ methods are all added with _ConfigureAwait(false)_. So I removed _await_.

      public Task<QueryResult<Dictionary<string, AgentService>>> Services(Filter filter, CancellationToken ct = default(CancellationToken))
        {
            return _client.Get<Dictionary<string, AgentService>>("/v1/agent/services", null, filter).Execute(ct);
        }

marcin-krystianc: Thanks @bosima for providing more context, now I see where are you coming from. I think that adding `ConfigureAwait(false)` is the way to go. 

Ps: I also think that we should probably use it in the whole library, not only in the places that are causing problems for you. Also we probably would want to use a Roslyn analyzer to enforce that. It is out of scope of your PR though, so don't worry about that.

bosima:
ok, i have changed the code.

marcin-krystianc (approved):
> ok, i have changed the code.

Thanks, looks good. 👍

marcin-krystianc:
@bosima We've released version v1.6.10.6 which contains your fix. Thank you for your contribution.

Issue #117, opened 2021-08-13
Registering a service in a namespace


ledjon-behluli:
Curious to know if there is any plan to support namespaces in Consul.NET ?
The namespaces functionality is available only in Consul Enterprise version 1.7.0 and later.

mfkl:
I don't see why not, anything from the Consul API should be considered. Would you like to try contribute a PR for this?

ledjon-behluli:
Sure! What's the best way to go about it? 
Should I fork branch development or master?

mfkl:
`master`, the develop branch is not used, we should remove it.

ledjon-behluli:
Got it!

winstxnhdw:
@marcin-krystianc We can close this issue now.

ledjon-behluli:
Yeah

PR #158, opened 2022-07-14
Switch to System.Text.Json

mfkl:
Rebased and iterated on https://github.com/G-Research/consuldotnet/pull/73

Few points to discuss:
- Bundled a few needed ConfigurationAwait(false) in the initial commit.
- which version of Text.Json do we want? [This comment](https://github.com/G-Research/consuldotnet/pull/73#discussion_r585898732) is on point but over a year passed, so we can probably pick a reasonable version now.
- [Marcin dynamic comment](https://github.com/G-Research/consuldotnet/pull/73#discussion_r585910845),
- do we want to keep a netcoreapp2.1 target? https://github.com/G-Research/consuldotnet/commit/f042debdb40692d3a8617f446f5b5d395553a3b5


mfkl:
If we want to bump the minor version to 1.7 for this change, we may want to provide bindings for all 1.7 Consul APIs, such as Namespaces https://github.com/G-Research/consuldotnet/issues/117. This needs a consul enterprise version, so I reached out to Consul for it. Let's see.

winstxnhdw:
We can probably continue with this.

marcin-krystianc:
> We can probably continue with this.

@winstxnhdw Feel free to give it a try.

Issue #159, opened 2022-07-27
[Bug]: Can Upgrade the Newtonsoft.Json to fix CWE-755?


Mutuduxf:
### Describe the bug

Introduced through
Consul@1.6.10.6
Fixed in
Newtonsoft.Json@13.0.1
Exploit maturity
PROOF OF CONCEPT
Show less detail 
Detailed paths
Introduced through: project@* › Consul@1.6.10.6 › Newtonsoft.Json@10.0.3
Fix: No remediation path available.
Overview
Affected versions of this package are vulnerable to Insecure Defaults due to improper handling of StackOverFlow exception (SOE) whenever nested expressions are being processed. Exploiting this vulnerability results in Denial Of Service (DoS), and it is exploitable when an attacker sends 5 requests that cause SOE in time frame of 5 minutes.

Note: This vulnerability affects IIS Applications.

### Steps To Reproduce

A security problem.

### Expected behavior

Upgrade the Newtonsoft.Json to 13.0.1 can fix it.

### Environment

- OS:       win 11
- Consul Version:   
- consultdotnet Version   consul1.6.10.6

PR #160, opened 2022-07-28
Update newtonsoftjson

mfkl:
See https://github.com/G-Research/consuldotnet/issues/159

### DIFF
--- a/Consul/Consul.csproj
+++ b/Consul/Consul.csproj
@@ -11,7 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


jgiannuzzi (approved):
Should we release this as another patch release or should we think about incrementing our version?
I know we still have a gap between the latest Consul API and what we actually support, but I wonder whether updating this dependency can be done with just a patch release.
@mfkl @marcin-krystianc thoughts?

marcin-krystianc (approved):
I'm fine with a patch release for bumping the version of `Newtonsoft.Json`. It is highly backwards compatible library, so the risk of accidental breakage for our users is very minimal.

PR #183, opened 2022-12-01
CI: use enterprise build

mfkl:
To use a consul enterprise version, 2 things are needed:
- An enterprise build (addressed by this PR),
- The consul enterprise key as an environment variable https://developer.hashicorp.com/nomad/tutorials/enterprise/hashicorp-enterprise-license. Someone with the necessary github rights will be able to do so.

I have tested the starting the enterprise agent locally.


Consul.Test/OperatorTest.cs
```diff
@@ -82,5 +82,12 @@ public async Task Operator_KeyringInstallListPutRemove()
                 Assert.True(response.Keys.ContainsKey(newKey));
             }
         }
+
+        [SkippableFact]
+        public async Task Operator_GetLicense()
+        {
+            var license = await _client.Operator.GetConsulLicense();
+            Skip.If(license.StatusCode == System.Net.HttpStatusCode.NotFound, "This test requires an Enterprise Consul version");
```
mfkl, line 90: This is not the most elegant, but it does the job. What do you guys think?

Consul Enterprise APIs, when called on the community edition server naturally return HTTP 404. 

With a XUnit `SkippableFact`, we can implement tests for these Enterprise version-only APIs and they can be skipped by adding this 2 line check in the beginning of any of these tests when the server is not Enterprise.

marcin-krystianc: Another possibility is to implement custom `Fact` attribute, e.g.:
```
    public class EnterpriseOnlyFactAttribute : FactAttribute
    {
        public EnterpriseOnlyFactAttribute()
        {
            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("RUN_CONSUL_ENTERPRISE_TESTS")))
            {
                Skip = "Ignored non-0enterprise servers";
            }
        }
    }
```
And the use:
```
[EnterpriseOnlyFactAttribute]
public async Task Operator_GetLicense()
{
    var license = await _client.Operator.GetConsulLicense();
    ...
}

It lets us to skip enterprise-only facts without making any extra http requests. 
Also, it feels more elegant in my opinion.

Comment (mfkl) [reply Consul.Test/OperatorTest.cs:90]: Much better, thanks!

File: .github/workflows/ci.yml [hunk 2]
```diff
@@ -84,12 +84,22 @@ jobs:
         uses: actions/setup-dotnet@v3
         with:
           dotnet-version: 6.0.x
+      - name: Setup Consul Enterprise URL
```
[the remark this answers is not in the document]
mfkl: > please remember that the secret is not passed to workflows that are triggered by a pull request from a fork

Ok, I think the github secret (e.g. enterprise license) will need some changes too :-) @jgiannuzzi

Consul/Operator.cs
```diff
@@ -229,6 +229,44 @@ public Task<WriteResult> KeyringUse(string key, WriteOptions q, CancellationToke
         {
             return _client.Put("/v1/operator/keyring", new KeyringRequest() { Key = key }, q).Execute(ct);
         }
+
+        public Task<QueryResult<ConsulLicense>> GetConsulLicense(string datacenter = "", CancellationToken ct = default)
```
marcin-krystianc, line 233: I think it would be good to add a comment explaining that this is enterprise-only feature

Consul/Operator.cs
```diff
@@ -229,6 +229,44 @@ public Task<WriteResult> KeyringUse(string key, WriteOptions q, CancellationToke
...
+        {
+            return _client.Get<ConsulLicense>("/v1/operator/license", new QueryOptions { Datacenter = datacenter }).Execute(ct);
+        }
+    }
+
+    public class ConsulLicense
```
marcin-krystianc, line 239: Same here

.github/workflows/ci.yml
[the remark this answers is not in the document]
marcin-krystianc: Can't we just run the enterprise tests when the workflow is not triggered from a fork (regardless of the branch it is triggered from?).

marcin-krystianc (approved):

marcin-krystianc:
Taking the liberty to merge and see whether the secret is correctly configured.

PR #187, opened 2022-12-09
Fix to correctly handle query TTL

gordonhopes:
We've been getting JSON deserialisation errors when doing client list queries


marcin-krystianc (approved):
@gordonhopes thanks for this PR.

PR #195, opened 2023-02-17
Can I add a field to support udp health check

Peterosh:


marcin-krystianc:
Hi @Peterosh, UDP checks seem to be a quite recent addition to consul (https://github.com/hashicorp/consul/pull/12722). Do you have any experience using them? I'm asking because I would like us to implement a test for it.

Peterosh:
Sorry. I just saw it in the official documentation.

marcin-krystianc:
I tried it locally and I was able to create a new check of `udp` type but I wasn't able to figure it out what to do to make the check to get healthy status. To my understanding, consul should periodically send udp datagrams to configured address. If it is successful (in case of UDP it means no ICMP message back) the check should go healthy. 
So I tried such test code:
``` csharp
[Fact]
public async Task Agent_Services_CheckUDP()
{
    var svcID = KVTest.GenerateTestKeyName();
    var check1Id = svcID + "_checkId";
    var registration = new AgentServiceRegistration
    {
        Name = svcID,
        Tags = new[] { "bar", "baz" },
        Port = 8000,
        Check = new AgentServiceCheck
        {
            CheckID = check1Id,
            UDP = "localhost:5000",
            Interval = TimeSpan.FromSeconds(10),
            Timeout = TimeSpan.FromSeconds(1),
        }
    };

    var services = await _client.Agent.Services();
    Assert.True(services.Response.ContainsKey(svcID));

    for (int i = 0; i < 10; i++)
    {
        var checks = await _client.Agent.Checks();
        var json = JsonConvert.SerializeObject(checks.Response[check1Id], Formatting.Indented);
        _testOutputHelper.WriteLine(json);
        _testOutputHelper.WriteLine("");
        await Task.Delay(TimeSpan.FromSeconds(5));
    }
}
```
and I've used ncat to open udp socket:
```
ncat -u -v localhost 5000
```

but it didn't work.

Peterosh:
I look at the documentation.The check status is set to success if any response is received from the targeted UDP server.Therefore, the udp listener needs to reply a data after receiving the message. [Look at this link](https://developer.hashicorp.com/consul/docs/discovery/checks#udp-check)

marcin-krystianc:
> I look at the documentation.The check status is set to success if any response is received from the targeted UDP server.Therefore, the udp listener needs to reply a data after receiving the message. [Look at this link](https://developer.hashicorp.com/consul/docs/discovery/checks#udp-check)

The problem is that I don't see any incoming traffic so there is nothing to respond to. Also in the consul agent logs there is absolutely nothing about querying the udp check.

Peterosh:
Well, it's true that I haven't tested it myself, I've just seen documentation that supports it

PR #219, opened 2023-06-20
Adding features

firerain-fd:
class Selectors:
Add Property ServiceName

class ServiceSelector:
Add operator ==
Add operator !=
Add Tests

class Agent:
Add method Checks with filter
Add Tests

### DIFF
--- a/Consul.Test/AgentTest.cs
+++ b/Consul.Test/AgentTest.cs
@@ -541,9 +541,11 @@ namespace Consul.Test
             var metaSelector = new MetaSelector();
 
             Assert.Equal(svcID1, (await _client.Agent.Services(idSelector == svcID1)).Response.Keys.Single());
-            Assert.Equal(svcID1, (await _client.Agent.Services(idSelector == svcID1)).Response.Keys.Single());
+            Assert.Equal(svcID2, (await _client.Agent.Services(idSelector == svcID2)).Response.Keys.Single());
             Assert.Equal(svcID1, (await _client.Agent.Services(metaSelector[uniqueMeta] == "bar1")).Response.Keys.Single());
             Assert.Equal(svcID2, (await _client.Agent.Services(metaSelector[uniqueMeta] == "bar2")).Response.Keys.Single());
+            Assert.Equal(svcID1, (await _client.Agent.Services(Selectors.Service == svcID1)).Response.Keys.Single());
+            Assert.Equal(svcID2, (await _client.Agent.Services(Selectors.Service == svcID2)).Response.Keys.Single());
         }
 
         [Theory]
--- a/Consul.Test/AgentTest.cs
+++ b/Consul.Test/AgentTest.cs
@@ -793,5 +793,42 @@ namespace Consul.Test
             Assert.Null(destinationProxyService.Proxy.Upstreams);
             Assert.Equal(ServiceKind.ConnectProxy, destinationProxyService.Kind);
         }
+
+        [Fact]
+        public async Task Agent_FilterChecks()
+        {
+            // Arrange
+            string svcName = KVTest.GenerateTestKeyName();
+            string svcID = $"{svcName}-1";
+            string checkName = $"Check {svcID}";
+
+            await _client.Agent.ServiceRegister(new AgentServiceRegistration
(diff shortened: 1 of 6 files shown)


Consul/Filtering/Selectors/ServiceSelector.cs
```diff
@@ -29,6 +29,9 @@ public sealed class ServiceSelector : Selector
         public MetaSelector Meta { get; } = new MetaSelector(Self);
 
         public override string Encode() => Self;
+
+        public static Filter operator ==(ServiceSelector selector, string value) => Filters.Eq(selector, value);
```
marcin-krystianc, line 33: I don't think a "service" is a valid selector that can be used for anything. Also, this is not being used in an actual test (apart from encoding tests). I think that it should be removed.

Consul/Interfaces/IAgentEndpoint.cs
```diff
@@ -33,6 +33,7 @@ public interface IAgentEndpoint
         Task<WriteResult> CheckDeregister(string checkID, CancellationToken ct = default(CancellationToken));
         Task<WriteResult> CheckRegister(AgentCheckRegistration check, CancellationToken ct = default(CancellationToken));
         Task<QueryResult<Dictionary<string, AgentCheck>>> Checks(CancellationToken ct = default(CancellationToken));
+        Task<QueryResult<Dictionary<string, AgentCheck>>> Checks(Filter filter, CancellationToken ct = default(CancellationToken));
```
marcin-krystianc, line 36: 👍

Consul.Test/AgentTest.cs
```diff
@@ -793,5 +793,42 @@ public async Task Agent_Service_Register_With_Connect()
             Assert.Null(destinationProxyService.Proxy.Upstreams);
             Assert.Equal(ServiceKind.ConnectProxy, destinationProxyService.Kind);
         }
+
+        [Fact]
+        public async Task Agent_FilterChecks()
```
marcin-krystianc, line 798: 👍

marcin-krystianc (commented):
Hi @firerain-fd, thanks for being an active contributor.

I've reviewed your changes. I see how the "ServiceName" selector is being used, but I don't see how "Service" selector is being used (or can be even used). Please provide a use case (a test case) or remove that change.

Consul/Filtering/Selectors/ServiceSelector.cs
firerain-fd: Hi.
The 'Service' selector is used in the following api queries:
https://developer.hashicorp.com/consul/api-docs/agent/service#list-services
https://developer.hashicorp.com/consul/api-docs/catalog#retrieve-map-of-services-for-a-node
https://developer.hashicorp.com/consul/api-docs/catalog#list-services-for-node

marcin-krystianc: Ok, I see. It would be great though if you implement at least one test case which uses this APIs.

Consul.Test/AgentTest.cs
```diff
@@ -541,9 +541,11 @@ public async Task Agent_FilterServices()
             var metaSelector = new MetaSelector();
 
             Assert.Equal(svcID1, (await _client.Agent.Services(idSelector == svcID1)).Response.Keys.Single());
-            Assert.Equal(svcID1, (await _client.Agent.Services(idSelector == svcID1)).Response.Keys.Single());
+            Assert.Equal(svcID2, (await _client.Agent.Services(idSelector == svcID2)).Response.Keys.Single());
             Assert.Equal(svcID1, (await _client.Agent.Services(metaSelector[uniqueMeta] == "bar1")).Response.Keys.Single());
             Assert.Equal(svcID2, (await _client.Agent.Services(metaSelector[uniqueMeta] == "bar2")).Response.Keys.Single());
+            Assert.Equal(svcID1, (await _client.Agent.Services(Selectors.Service == svcID1)).Response.Keys.Single());
```
marcin-krystianc, line 547: 👍

marcin-krystianc (approved):
Nice, thank you for your contribution!

PR #227, opened 2023-07-24
Unblock the CI for enterprise version

marcin-krystianc:
This PR:
1. Fixes version parsing by using a class that properly understands semantic versioning (e.g. version `1.6.10+ent`)
2. We run Enterprise tests on all branches. 
3. We don't have the license environment variable set yet, but it turns out that older versions of Consul can be run anyway.


marcin-krystianc:
This is ready for review.

mfkl (approved):
> We don't have the license environment variable set yet

I guess this would be for a follow up PR then (and whether we want to enable the secret for PR builds).

PR #248, opened 2023-10-19
Added Support for Retrieving Nodes that support Mesh Capable Service

sammychinedu2ky:
This pull request adds support for this 
https://developer.hashicorp.com/consul/api-docs/catalog#list-nodes-for-mesh-capable-service

### DIFF
--- a/Consul/Catalog.cs
+++ b/Consul/Catalog.cs
@@ -243,7 +243,7 @@ namespace Consul
         /// </summary>
         /// <param name="service">The service ID</param>
         /// <param name="ct">Cancellation token for long poll request. If set, OperationCanceledException will be thrown if the request is cancelled before completing</param>
-        /// <returns></returns>
+        /// <returns>A list of service instances</returns>
         public Task<QueryResult<CatalogService[]>> NodesForMeshCapableService(string service, CancellationToken ct = default)
         {
             return NodesForMeshCapableService(service, QueryOptions.Default, null, ct);
--- a/Consul/Catalog.cs
+++ b/Consul/Catalog.cs
@@ -246,7 +246,7 @@ namespace Consul
         /// <returns></returns>
         public Task<QueryResult<CatalogService[]>> NodesForMeshCapableService(string service, CancellationToken ct = default)
         {
-            return NodesForMeshCapableService(service,QueryOptions.Default, null, ct);
+            return NodesForMeshCapableService(service, QueryOptions.Default, null, ct);
         }
 
         /// <summary>
--- a/Consul/Catalog.cs
+++ b/Consul/Catalog.cs
@@ -268,7 +268,7 @@ namespace Consul
         /// <param name="q">Customized Query options</param>
         /// <param name="filter">Specifies the expression used to filter the queries results prior to returning the data</param>
         /// <param name="ct">Cancellation token for long poll request. If set, OperationCanceledException will be thrown if the request is cancelled before completing</param>
-        /// <returns></returns>
+        /// <returns>A list of service instances</returns>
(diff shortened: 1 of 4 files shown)


Consul/Catalog.cs
```diff
@@ -237,6 +238,44 @@ public Task<QueryResult<CatalogService[]>> Service(string service, string tag, Q
             return req.Execute(ct);
         }
 
+        /// <summary>
+        /// Returns the nodes providing a mesh-capable service in a given datacenter. 
+        /// </summary>
+        /// <param name="service">The service ID</param>
+        /// <param name="ct">Cancellation token for long poll request. If set, OperationCanceledException will be thrown if the request is cancelled before completing</param>
+        /// <returns></returns>
```
winstxnhdw, line 246:
```suggestion
        /// <returns>A list of service instances</returns>
```

Consul/Catalog.cs
```diff
@@ -237,6 +238,44 @@ public Task<QueryResult<CatalogService[]>> Service(string service, string tag, Q
...
+            return NodesForMeshCapableService(service, QueryOptions.Default, null, ct);
+        }
+
+        /// <summary>
+        /// Returns the nodes providing a mesh-capable service in a given datacenter.
+        /// </summary>
+        /// <param name="service">The service ID</param>
+        /// <param name="filter">Specifies the expression used to filter the queries results prior to returning the data</param>
+        /// <param name="ct">Cancellation token for long poll request. If set, OperationCanceledException will be thrown if the request is cancelled before completing</param>
+        /// <returns></returns>
```
winstxnhdw, line 258:
```suggestion
        /// <returns>A list of service instances</returns>
```

Consul/Catalog.cs
```diff
@@ -237,6 +238,44 @@ public Task<QueryResult<CatalogService[]>> Service(string service, string tag, Q
...
+        }
+
+        /// <summary>
+        /// Returns the nodes providing a mesh-capable service in a given datacenter.
+        /// </summary>
+        /// <param name="service">The service ID</param>
+        /// <param name="q">Customized Query options</param>
+        /// <param name="filter">Specifies the expression used to filter the queries results prior to returning the data</param>
+        /// <param name="ct">Cancellation token for long poll request. If set, OperationCanceledException will be thrown if the request is cancelled before completing</param>
+        /// <returns></returns>
```
winstxnhdw, line 271:
```suggestion
        /// <returns>A list of service instances</returns>
```

winstxnhdw (commented):
I didn't actually review it but I think these should be filled up :D

Consul/Catalog.cs
sammychinedu2ky: thanks man

marcin-krystianc (commented):
Hi @sammychinedu2ky,

the implementation seems very reasonable, but I dislike that we cannot create mesh-capable services with our library. Without it, the test scenario is kind of artificial, because we are not testing whether the endpoint actually returns anything.
Would you want to investigate more on mesh-capable services and what changes we would need to be able to create them with our library?

sammychinedu2ky:
> Hi @sammychinedu2ky,
> 
> the implementation seems very reasonable, but I dislike that we cannot create mesh-capable services with our library. Without it, the test scenario is kind of artificial, because we are not testing whether the endpoint actually returns anything. Would you want to investigate more on mesh-capable services and what changes we would need to be able to create them with our library?

Alright Chief

sammychinedu2ky:
The mesh service focuses more on Kubernetes support, which the test environment doesn't create

sammychinedu2ky:
@marcin-krystianc .. please can you review this one first

sammychinedu2ky:
Resolved by https://github.com/G-Research/consuldotnet/pull/273

PR #267, opened 2023-11-08
CI: Add DependaBot checks

ljubon:
The goal of this PR is to add DependaBot workflow 
https://github.com/G-Research/oss-portfolio-maturity/issues/278

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,28 @@
+---
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "nuget"
+    directory: "/Consul/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,28 @@
+---
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
```
marcin-krystianc, line 5: Hi @ljubon,

similarly to the [ParquetSharp.DataFrame](https://github.com/G-Research/ParquetSharp.DataFrame/pull/10#discussion_r1406806189), consuldotnet is a library for which we care about backwards compatibility. So I think it makes sense to leave the dependabot only for `github-actions`.

marcin-krystianc (approved):

marcin-krystianc (approved):

Issue #276, opened 2023-11-30
Support Getting Service Configuration


sammychinedu2ky:
### Is your feature request related to a problem? Please describe.

no

### Describe the solution you'd like

add support for (https://developer.hashicorp.com/consul/api-docs/agent/service#get-service-configuration)
This API is supported in the golang client (1.8.19)

PR #278, opened 2023-12-01
Added Support for Getting Service Configuration

sammychinedu2ky:
Fixes https://github.com/G-Research/consuldotnet/issues/276
@marcin-krystianc


marcin-krystianc (dismissed):

marcin-krystianc (commented):
I think it would be more consistent with the rest of the system, to use `ServiceId` capitalisation.
I know that we have form ``ServiceID` in some classes, but as a parameter in a method `ServiceId` form seems more suitable.

sammychinedu2ky:
done

marcin-krystianc (approved):

PR #298, opened 2024-01-25
Add the Configuration property to the IConsulClient interface

firerain-fd:
The Configuration endpoint cannot be used in DI because the Configuration property is not declared in the IConsulClient interface.


marcin-krystianc (approved):

marcin-krystianc:
@firerain-fd, thank you for your contribution.

marcin-krystianc:
It should be available for use now -> https://www.nuget.org/packages/Consul/1.7.14.3

Issue #311, opened 2024-04-05
FEATURE REQUEST: Support for List Nodes for Connect-capable Service


JocelynVelarde:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Add support to [List Nodes for Connect-capable Service](https://developer.hashicorp.com/consul/api-docs/v1.9.x/health#list-nodes-for-connect-capable-service) 
This endpoint should return the nodes providing a Connect-capable service in a given datacenter. This will include both proxies and native integrations. 
This API is supported in the Golang client ([1.9.17](https://github.com/hashicorp/consul/blob/4897b16abf192942d1e1d2ffa109bf17b2431e9a/api/health.go#L311))

### Describe alternatives you've considered

Nothing for the moment

### Additional context

Nothing for the moment

JocelynVelarde:
- Merged in #312

PR #312, opened 2024-04-05
Added list nodes for connect-capable service with endpoints

JocelynVelarde:
## Changes made
1. Added GET /health/connect/:service into `Health.cs`
2. Included new endpoints into `IHealthEndpoint.cs`

Refers to issue #311 

## Next tasks
- [ ] Write corresponding tests

### DIFF
--- a/Consul.Test/HealthTest.cs
+++ b/Consul.Test/HealthTest.cs
@@ -21,6 +21,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Xunit;
+using Consul.Filtering; 
 
 namespace Consul.Test
 {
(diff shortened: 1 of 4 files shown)


JocelynVelarde:
Hello! I would appreciate some feedback/review regarding this new endpoint support. 🐝

Consul/Health.cs
```diff
@@ -284,11 +298,22 @@ public Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, boo
...
+            switch (healthType)
+            {
+                case "connectHealth":
+                    path = "/v1/health/connect/{0}";
+                    break;
+                default:
+                    path = "/v1/health/service/{0}";
+                    break;
+            }
+            var req = _client.Get<ServiceEntry[]>(string.Format(path, service), q, filter);
```
marcin-krystianc, line 316: I know that the documentation mentions that the `connect` and `service` endpoints are nearly identical, but I don't think it's a good idea to rely on that. I would rather split this into two separate implementations, one for `connect` and one for `service` endpoints.

Consul/Interfaces/IHealthEndpoint.cs
```diff
@@ -36,7 +36,9 @@ public interface IHealthEndpoint
         Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, CancellationToken ct = default);
         Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, bool passingOnly, CancellationToken ct = default);
         Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, bool passingOnly, QueryOptions q, CancellationToken ct = default);
-        Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, bool passingOnly, QueryOptions q, Filter filter, CancellationToken ct = default);
+        Task<QueryResult<ServiceEntry[]>> Service(string service, string tag, bool passingOnly, QueryOptions q, Filter filter, string healthType, CancellationToken ct = default);
```
marcin-krystianc, line 39: It is not ok, to change the public method signature in a way that makes it incompatible with previous versions. 
If a new parameter is needed, then the new method overload needs to be added (without modifying the existing one).

marcin-krystianc (commented):
Hi @JocelynVelarde,

Thank you for raising this PR. I've left a comment regarding the importance of maintaining backward compatibility and avoiding unnecessary logic in our library. Apart from that, could you please implement tests for the new API that is being introduced?

Consul/Health.cs
JocelynVelarde: Okei noted, so instead of implementing the switch case, I should implement as a separate entity?

Consul/Interfaces/IHealthEndpoint.cs
JocelynVelarde: Alright so, creating leave the method signature as it was and just add a new one with the new parameter.

Consul/Health.cs
marcin-krystianc: > Okei noted, so instead of implementing the switch case, I should implement as a separate entity?

yes, I think it is going to be better that way.

JocelynVelarde:
Hello! I was getting some errors with the `BaseFixture.cs` class, I was wondering if someone could explain why this is happening, greetings.

marcin-krystianc:
> Hello! I was getting some errors with the `BaseFixture.cs` class, I was wondering if someone could explain why this is happening, greetings.

Hi, I'm not sure which errors you mean exactly. I see that everything fails because the response is empty. I suggest have a look at `Agent_Service_Register_With_Connect` and see how to register service with a "connect" information.

JocelynVelarde:
> > Hello! I was getting some errors with the `BaseFixture.cs` class, I was wondering if someone could explain why this is happening, greetings.
> 
> Hi, I'm not sure which errors you mean exactly. I see that everything fails because the response is empty. I suggest have a look at `Agent_Service_Register_With_Connect` and see how to register service with a "connect" information.

From what I can see, Connect is equivalent to Service, except that Service will return the ones that are Connect-enabled. The new endpoint for Connect in `Health.cs` is correct? I just noticed that there is a `Filter` parameter that I was not using in the test request. Here
`var checks = await _client.Health.Connect(svcID, "", false, QueryOptions.Default);`
I'm not sure if this will fix the error.

marcin-krystianc:
> > > Hello! I was getting some errors with the `BaseFixture.cs` class, I was wondering if someone could explain why this is happening, greetings.
> > 
> > 
> > Hi, I'm not sure which errors you mean exactly. I see that everything fails because the response is empty. I suggest have a look at `Agent_Service_Register_With_Connect` and see how to register service with a "connect" information.
> 
> From what I can see, Connect is equivalent to Service, except that Service will return the ones that are Connect-enabled. The new endpoint for Connect in `Health.cs` is correct? I just noticed that there is a `Filter` parameter that I was not using in the test request. Here `var checks = await _client.Health.Connect(svcID, "", false, QueryOptions.Default);` I'm not sure if this will fix the error.

I think, the key to properly test it is to create conenct-enabled service, e.g.:
```
            var destinationServiceRegistrationParameters = new AgentServiceRegistration
            {
                ID = destinationServiceID,
                Name = destinationServiceID,
                Port = 8000,
                Check = new AgentServiceCheck
                {
                    TTL = TimeSpan.FromSeconds(15)
                },
                Connect = new AgentServiceConnect
                {
                    SidecarService = new AgentServiceRegistration
                    {
                        Port = 8001
                    }
                }
            };
```

marcin-krystianc (approved):
Hi @JocelynVelarde,

the changes look good now. and the tests are successful in the CI too. There is a small issue with GH reporting status to the RP though. Can you push another commit or two (e.g. add an empty line and remove it) maybe that will resolve the issue with tests statuses?

marcin-krystianc:
> Hi @JocelynVelarde,
> 
> the changes look good now. and the tests are successful in the CI too. There is a small issue with GH reporting status to the RP though. Can you push another commit or two (e.g. add an empty line and remove it) maybe that will resolve the issue with tests statuses?

Nevermind, I closed and reopened the PR to trigger the workflow on the latest commit.

Consul.Test/HealthTest.cs
```diff
@@ -116,6 +117,50 @@ private struct AggregatedStatusResult
...
+            finally
+            {
+                await _client.Agent.ServiceDeregister(destinationServiceID);
+            }
+        }
+
+
+
+
+
```
marcin-krystianc, line 163: nit: ```suggestion
```

marcin-krystianc:
I think that the `Catalog_GetNodesForMeshCapableService` got flaky after adding the `Health_Connect' test.
We need to keep eye on it, if it is going to fail frequently, then we will need to investigate the root cause.

marcin-krystianc (approved):

Issue #313, opened 2024-04-10
FEATURE REQUEST: Read LAN Coordinates for a node


JocelynVelarde:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Add support to [Read LAN Coordinates for a node](https://developer.hashicorp.com/consul/api-docs/v1.9.x/coordinate#read-lan-coordinates-for-a-node) 
This endpoint returns the LAN network coordinates for the given node.
This API is supported in the Golang client ([1.9.17](https://github.com/hashicorp/consul/blob/4897b16abf192942d1e1d2ffa109bf17b2431e9a/api/coordinate.go#L88))

### Describe alternatives you've considered

Nothing for the moment

### Additional context

Nothing for the moment

JocelynVelarde:
- Merged in #314

PR #314, opened 2024-04-11
Read LAN Coordinates for a specific node

JocelynVelarde:
## Changes made
1. Added GET /v1/coordinate/node/+node into `Coordinate.cs`
2. Included new endpoints into `ICoordinateEndpoint.cs`
3. Added test `CoordinateTest.cs`

Refers to issue #313

### DIFF
--- a/Consul.Test/BaseFixture.cs
+++ b/Consul.Test/BaseFixture.cs
@@ -22,7 +22,7 @@ namespace Consul.Test
                 c.Address = TestHelper.HttpUri;
             });
 
-            var timeout = TimeSpan.FromSeconds(15);
+            var timeout = TimeSpan.FromSeconds(60);
             var cancelToken = new CancellationTokenSource(timeout).Token;
             Exception exception = null;
             var firstIteration = true;
@@ -66,6 +66,9 @@ namespace Consul.Test
 
                     // Workaround for https://github.com/hashicorp/consul/issues/15061
                     await client.Agent.GetAgentMetrics();
+
+                    if(await client.Coordinate.Nodes()).Response.Length == 0)
+                        continue;
 
                     break;
                 }
--- a/Consul.Test/BaseFixture.cs
+++ b/Consul.Test/BaseFixture.cs
@@ -67,8 +67,8 @@ namespace Consul.Test
                     // Workaround for https://github.com/hashicorp/consul/issues/15061
                     await client.Agent.GetAgentMetrics();
 
-                    if((await client.Coordinate.Nodes()).Response.Length == 0)
-            continue;
+                    if ((await client.Coordinate.Nodes()).Response.Length == 0)
+                        continue;
 
                     break;
                 }
--- a/Consul.Test/BaseFixture.cs
+++ b/Consul.Test/BaseFixture.cs
@@ -22,7 +22,7 @@ namespace Consul.Test
             c.Address = TestHelper.HttpUri;
         });
 
-        var timeout = TimeSpan.FromSeconds(60);
+        var timeout = TimeSpan.FromSeconds(15);
         var cancelToken = new CancellationTokenSource(timeout).Token;
         Exception exception = null;
         var firstIteration = true;
@@ -67,8 +67,10 @@ namespace Consul.Test
                 // Workaround for https://github.com/hashicorp/consul/issues/15061
                 await client.Agent.GetAgentMetrics();
 
-                    if ((await client.Coordinate.Nodes()).Response.Length == 0)
-                        continue;
+                    /*var nodesResponse = await client.Coordinate.Nodes();
+                    if (nodesResponse != null && nodesResponse.Response.Length == 0)
+                        continue;*/
+
                     break;
                 }
                 catch (OperationCanceledException)
--- a/Consul.Test/BaseFixture.cs
+++ b/Consul.Test/BaseFixture.cs
@@ -16,56 +16,56 @@ namespace Consul.Test
 
         private static readonly Lazy<Task> Ready = new Lazy<Task>(async () =>
         {
-        var client = new ConsulClient(c =>
-        {
-            c.Token = TestHelper.MasterToken;
-            c.Address = TestHelper.HttpUri;
-        });
-
-        var timeout = TimeSpan.FromSeconds(15);
-        var cancelToken = new CancellationTokenSource(timeout).Token;
-        Exception exception = null;
-        var firstIteration = true;
-        while (true)
-        {
-            if (!firstIteration)
-                await Task.Delay(TimeSpan.FromSeconds(1), cancelToken);
-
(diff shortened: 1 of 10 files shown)


Consul/Interfaces/ICoordinateEndpoint.cs
```diff
@@ -29,6 +29,7 @@ public interface ICoordinateEndpoint
     {
         Task<QueryResult<CoordinateDatacenterMap[]>> Datacenters(CancellationToken ct = default);
         Task<QueryResult<CoordinateEntry[]>> Nodes(CancellationToken ct = default);
+        Task<QueryResult<CoordinateEntry>> Node(string node, QueryOptions q, CancellationToken ct = default);
```
marcin-krystianc, line 32: There are two overloads of `Nodes` endpoint (with and without `QueryOptions q` parameter).
We can Apply the same pattern for the new `Node` endpoint`

marcin-krystianc (commented):
Tests are failing to compile (because of missing query options parameter). 
Please note also that the 'Node' endpoint returns an array (https://developer.hashicorp.com/consul/api-docs/v1.9.x/coordinate#sample-response-2), instead of a single object.
I'm not sure how to handle it. It seems unlikely that there will be more than one entry in the result, but since the endpoint returns an array we probably should use an array in the result as well.

Consul/Interfaces/ICoordinateEndpoint.cs
JocelynVelarde: Sure noted! Already fixed 🚀

JocelynVelarde:
> Tests are failing to compile (because of missing query options parameter). Please note also that the 'Node' endpoint returns an array (https://developer.hashicorp.com/consul/api-docs/v1.9.x/coordinate#sample-response-2), instead of a single object. I'm not sure how to handle it. It seems unlikely that there will be more than one entry in the result, but since the endpoint returns an array we probably should use an array in the result as well.

I already fixed this on the recent commit, let me know if it needs further adjustments :)

marcin-krystianc:
> > Tests are failing to compile (because of missing query options parameter). Please note also that the 'Node' endpoint returns an array (https://developer.hashicorp.com/consul/api-docs/v1.9.x/coordinate#sample-response-2), instead of a single object. I'm not sure how to handle it. It seems unlikely that there will be more than one entry in the result, but since the endpoint returns an array we probably should use an array in the result as well.
> 
> I already fixed this on the recent commit, let me know if it needs further adjustments :)

I've run the CI and there are some compilation issues. Please have a look.

marcin-krystianc:
Hi @JocelynVelarde,

there is a problem with the `coordinates` endpoint, which returns empty responses for some time after starting the consul agent (e.g. https://github.com/G-Research/consuldotnet/actions/runs/8692709325/job/23862884528).
We can fix this by bumping the wait time and adding another check in the Base Fixture class:
image

JocelynVelarde:
> Hi @JocelynVelarde,
> 
> 
> 
> there is a problem with the `coordinates` endpoint, which returns empty responses for some time after starting the consul agent (e.g. https://github.com/G-Research/consuldotnet/actions/runs/8692709325/job/23862884528).
> 
> We can fix this by bumping the wait time and adding another check in the Base Fixture class:
> 
> image
> 
> 
> 
> 

Alright noted! Is there any way I can test the endpoint locally (making the request) so I can check the output before submitting for review?

marcin-krystianc:
> 
> Alright noted! Is there any way I can test the endpoint locally (making the request) so I can check the output before submitting for review?

Sure thing, there is a description how to to start local agent for tests -> https://consuldot.net/docs/getting-started/setting-up-server

JocelynVelarde:
> 
> 
> > 
> 
> > Alright noted! Is there any way I can test the endpoint locally (making the request) so I can check the output before submitting for review?
> 
> 
> 
> Sure thing, there is a description how to to start local agent for tests -> https://consuldot.net/docs/getting-started/setting-up-server

Yeaaa! That's how I started running tests yesterday, so on the CLI I have to write the fetch request?

Consul/Coordinate.cs
```diff
@@ -88,6 +88,23 @@ public Task<QueryResult<CoordinateEntry[]>> Nodes(QueryOptions q, CancellationTo
...
+        /// <param name="q">Customized query options</param>"
+        /// <remarks>Node is used to return the coordinates of a given node in the LAN pool.</returns>
+        public Task<QueryResult<CoordinateEntry[]>> Node(string node, QueryOptions q, CancellationToken ct = default)
+        {
+            return _client.Get<CoordinateEntry[]>(string.Format("/v1/coordinate/node/{0}", node), q).Execute(ct);
+        }
+
+        public Task<QueryResult<CoordinateEntry[]>> Node(string node, CancellationToken ct = default)
+        {
+            return _client.Get<CoordinateEntry[]>(string.Format("/v1/coordinate/node/{0}", node)).Execute(ct);
```
marcin-krystianc, line 106: Can be replaced with:
```suggestion
            return Node(node, QueryOptions.Default, ct);
```

Consul.Test/BaseFixture.cs
```diff
@@ -67,6 +67,10 @@ public class BaseFixture : IAsyncLifetime
                     // Workaround for https://github.com/hashicorp/consul/issues/15061
                     await client.Agent.GetAgentMetrics();
 
+                    /*var nodesResponse = await client.Coordinate.Nodes();
```
marcin-krystianc, line 70: Is it commented out by mistake?

Consul.Test/CoordinateTest.cs
```diff
@@ -49,5 +49,19 @@ public async Task Coordinate_GetNodes()
             // get an error. - from offical API.
             Assert.NotNull(nodes);
         }
+
+        [Fact]
+        public async Task Coordinate_GetNode()
+        {
```
marcin-krystianc, line 55: Previous version of the test was better, we did actually use a real node name in the query before.

Consul.Test/BaseFixture.cs
JocelynVelarde: Ohh let me activate it again

Consul/Coordinate.cs
JocelynVelarde: Alright done

Consul.Test/CoordinateTest.cs
JocelynVelarde: The thing is that with the node name I was receiving errors in the base fixture class

marcin-krystianc (commented):
Thanks, looks good. Just one final remark about removing unnecessary code.

marcin-krystianc (approved):
Now it is perfect, thank you!
I need to sort out macOS CI failures before merging it.

Issue #315, opened 2024-04-11
FEATURE REQUEST: List Services for Gateway


JocelynVelarde:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Add support to [List Services for Gateway](https://developer.hashicorp.com/consul/api-docs/v1.9.x/catalog#list-services-for-gateway) 
This endpoint returns the services associated with an ingress gateway or terminating gateway.
This API is supported in the Golang client ([1.9.17](https://github.com/hashicorp/consul/blob/4897b16abf192942d1e1d2ffa109bf17b2431e9a/api/catalog.go#L310))

### Describe alternatives you've considered

Nothing for the moment

### Additional context

Nothing for the moment

IvanKolchanov:
Hi! My name is Ivan Kolchanov, I am a MLH fellow working on Consul.NET right now. I have been working on features to get Consul.NETup-to-date with 1.9.17. Just wanted to ask if you are going to finish GET Gateaway services any time soon? If you are busy and do not have time for the project now I can finish that endpoint. Thank you in advance!

JocelynVelarde:
> Hi! My name is Ivan Kolchanov, I am a MLH fellow working on Consul.NET right now. I have been working on features to get Consul.NETup-to-date with 1.9.17. Just wanted to ask if you are going to finish GET Gateaway services any time soon? If you are busy and do not have time for the project now I can finish that endpoint. Thank you in advance!

Hello! Glad to see your contributions. Yes of course, you can go ahead and finish the issues. I believe this specific one, was almost ready to merge, there where just some unexpected errors. If you wish you can check the PR conversation to see what is done and what is there to fix.

PR #321, opened 2024-05-08
List services gateway

JocelynVelarde:
## Changes made
1. Added GET /v1/catalog/gateway-services/+gateway into `Catalog.cs`
2. Included new endpoint into `ICatalogEndpoint.cs`
3. Added possible test `CatalogTest.cs` (currently as a comment)

- Refers to issue #315


JocelynVelarde:
Hello greetings! ☀️ 
I adapted the commented test, basing myself on the already implemented one from the Consul codebase. Still, I can see that I have to create a `TerminatingGatewayConfigEntry` struct. Is this necessary? Also, I was wondering what is the difference between `Gateaway_ingress` and `Gateaway_terminating`

marcin-krystianc:
> Gateaway_terminating

Hi,

I'm unsure, but perhaps there's no need to create a separate TerminatingGatewayConfigEntry. It might be possible to utilize the AgentService struct and assign the relevant ServiceKind value accordingly. 
As for the difference between ingress and terminating gateways, please refer to the documentation (https://developer.hashicorp.com/consul/docs/connect/gateways). 

For our testing purposes, registering any ingress and terminating gateway and then querying it should suffice.

JocelynVelarde:
> > Gateaway_terminating
> 
> Hi,
> 
> I'm unsure, but perhaps there's no need to create a separate TerminatingGatewayConfigEntry. It might be possible to utilize the AgentService struct and assign the relevant ServiceKind value accordingly. As for the difference between ingress and terminating gateways, please refer to the documentation (https://developer.hashicorp.com/consul/docs/connect/gateways).
> 
> For our testing purposes, registering any ingress and terminating gateway and then querying it should suffice.

OH great, would it look something like this? 

```
 var terminatingGatewayService = new AgentService
    {
        ID = svcID + "-terminating",
        Service = svcID,
        Tags = new[] { "gateway", "terminating" },
        Port = 9090,
        Kind = ServiceKind.TerminatingGateway, 
    };
```

marcin-krystianc:
I guess it might work, worth giving it a try.

JocelynVelarde:
@marcin-krystianc I believe it is working now! 

`2 failing, 222 successful, and 3 skipped checks` → on GitHub CLI

Consul.Test/CatalogTest.cs
```diff
@@ -306,15 +390,74 @@ public async Task Catalog_GatewayServices()
...
 
-                var terminatingGatewayServices = await client.Catalog.GatewayService("terminating-gateway", QueryOptions.Default, CancellationToken.None);
-                Assert.NotEmpty(terminatingGatewayServices.Response);
+                // Retrieve gateway services associated with the "gateway-service" node
+                var gatewayServices = await client.Catalog.GatewayService("gateway-service", QueryOptions.Default, CancellationToken.None);
 
-                var ingressGatewayServices = await client.Catalog.GatewayService("ingress-gateway", QueryOptions.Default, CancellationToken.None);
-                Assert.NotEmpty(ingressGatewayServices.Response);
+                // Assert that the response is not null
+                Assert.NotNull(gatewayServices);
```
marcin-krystianc, line 312: I think, it "works" because we don't actually check whether there are any services returned, it should be `Assert.NotEmpty(gatewayServices.Response)` 🐞

JocelynVelarde: Alright let me check. 🫡

IvanKolchanov:
@marcin-krystianc Hi, Marcin! I have picked up this issue for now and I just wanted to check up with you on what the state of the issue last was. I have been trying the code and for now it seems to me that during the test the gateaways fail to register completely or fail to register in time for when we check them. That is because when I run the tests the Catalog.Register requests return OK as well as /catalog/gateaway-services/..., however, the GateawayServices Response is Empty.

marcin-krystianc:
> @marcin-krystianc Hi, Marcin! I have picked up this issue for now and I just wanted to check up with you on what the state of the issue last was. I have been trying the code and for now it seems to me that during the test the gateaways fail to register completely or fail to register in time for when we check them. That is because when I run the tests the Catalog.Register requests return OK as well as /catalog/gateaway-services/..., however, the GateawayServices Response is Empty.

Yeah, as far I remember we couldn't figure out why the test procedure fails. It seems the registration is successful, i.e. the request returns Ok and in the Web UI (http://localhost:8500/) we could see it. But still the response is empty and the test fails. 
I think it is necessary to reach the consul community for some help.. Another option would be maybe trying to go's tests and seeing how they work.

IvanKolchanov:
> Yeah, as far I remember we couldn't figure out why the test procedure fails. It seems the registration is successful, i.e. the request returns Ok and in the Web UI (http://localhost:8500/) we could see it. But still the response is empty and the test fails. I think it is necessary to reach the consul community for some help.. Another option would be maybe trying to go's tests and seeing how they work.

Hmm, makes sense. How would we go about reaching out to the community? And I am not really sure if I understand the last part corrently. Do you mean trying to work out what we have different from the tests in the Golang client?

marcin-krystianc:
> Hmm, makes sense. How would we go about reaching out to the community? And I am not really sure if I understand the last part corrently. Do you mean trying to work out what we have different from the tests in the Golang client?

Regarding the community: I think Jocelyn already started a conversation in the consul project https://discuss.hashicorp.com/t/null-response-when-making-get-request-to-list-services-associated-with-a-gateway/67115/3.

Regarding the go tests, I mean that we can compare our tests with go client's tests and look for differences.

marcin-krystianc:
Replaced with https://github.com/G-Research/consuldotnet/pull/347

PR #324, opened 2024-05-12
#323 Add net 6.0

### LINKED ISSUE #323: Add additional TFM So Microsoft.CSharp can be externalised
**Is your feature request related to a problem? Please describe.**
I want to minimise dependencies in my project by utilising framework dependencies wherever possible

**Describe the solution you'd like**
I want the package to have an additional TFM (net standard 2.1) added so that the explicit dependency on Microsoft.CSharp  can be removed. 

**Describe alternatives you've considered**
Accept the additional dependency

**Additional context**
n/a

thompson-tomo:
Adds in net 6.0 so that Microsoft.CSharp is no longer required as a dependency and improve cancellation propagation

Closes #323

### DIFF
--- a/Consul/Consul.csproj
+++ b/Consul/Consul.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461;net6</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461;net6.0</TargetFrameworks>
     <Authors>PlayFab,$(Authors)</Authors>
     <Copyright>Copyright 2015-2018 PlayFab Inc.; $(Copyright)</Copyright>
     <Description>Consul.NET is a .NET client library for the Consul HTTP API</Description>
--- a/Consul/Consul.csproj
+++ b/Consul/Consul.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461;net6</TargetFrameworks>
     <Authors>PlayFab,$(Authors)</Authors>
     <Copyright>Copyright 2015-2018 PlayFab Inc.; $(Copyright)</Copyright>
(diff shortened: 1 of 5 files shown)


Consul/Consul.csproj
```diff
@@ -1,7 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461;netstandard2.1</TargetFrameworks>
```
marcin-krystianc, line 4: It seems that `netstandard2.1` is not implementing enough of `dynamic` type support to completely remove the "Microsoft.CSharp" dependency, I think it would work with `net6.0` instead of `netstandard2.1`. 
I wonder though, if it is worth adding a new target framework, just to kill one NuGet dependency?

thompson-tomo: have added net 6 instead and at the same time improved the cancellation toke propagation.. Hence closing conversation.

Consul/Consul.csproj
```diff
@@ -1,7 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461;net6</TargetFrameworks>
```
sungam3r, line 4:
```suggestion
    <TargetFrameworks>netstandard2.0;net461;net6.0</TargetFrameworks>
```

marcin-krystianc (commented):
Thanks @thompson-tomo for your patience and apologies for the delay with this PR.

Initially, I was hesitant about this change because our test matrix in the CI wasn't ready for adding a new target framework. Also, adding a new target framework to the library expands the API surface of the library that we need to make sure is tested.

But we've decided now that adding new tfms is inevitable, so we are going to move forward with this change.


- Since `.net6.0` is already out of support, should we update this PR to use net8.0 instead (current LTS)?

- Please add an older target framework (I guess the previous LTS, which is `net6.0`) to the test project so we are able to run tests for `netstandard2.0`.

- Please update the CI matrix so we run tests for `net461`, `net.6.0`, and `net8.0`.

Issue #331, opened 2024-06-05
[FEATURE REQUEST]: Connect CA Roots, return root certificates in the cluster


IvanKolchanov:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Add support to https://developer.hashicorp.com/consul/api-docs/v1.9.x/connect/ca
This endpoint returns the current list of trusted CA root certificates in the cluster.
This endpoint is supported in Golang client (1.9.17) https://github.com/hashicorp/consul/blob/v1.9.17/api/connect_ca.go#L134

### Describe alternatives you've considered

Nothing for the moment

### Additional context

Nothing for the moment

IvanKolchanov:
Resolved in #332

PR #332, opened 2024-06-11
Connect CA Roots

IvanKolchanov:
# Changes made
1. Added GET /v1/agent/connect/ca/roots endpoint
2. Included new endpoint into ICatalogEndpoint.cs
3. Added possible test to AgentTest.cs
4. Edited test_config.json for the test

# Issue ticket number and link
*  Refers to issue https://github.com/G-Research/consuldotnet/issues/331

### DIFF
--- a/Consul.Test/AgentTest.cs
+++ b/Consul.Test/AgentTest.cs
@@ -1036,7 +1036,6 @@ namespace Consul.Test
             var root = caRoots.Response.Roots.First();
             Assert.NotNull(root.ID);
             Assert.NotNull(root.Name);
-            Assert.NotEqual(0, root.SerialNumber);
             Assert.NotNull(root.SigningKeyID);
             Assert.NotNull(root.ExternalTrustDomain);
             Assert.NotNull(root.NotBefore);
@@ -1046,8 +1045,11 @@ namespace Consul.Test
             Assert.True(root.Active);
             Assert.NotNull(root.PrivateKeyType);
             Assert.NotEqual(0, root.PrivateKeyBits);
-            Assert.NotEqual(0, root.CreateIndex);
-            Assert.NotEqual(0, root.ModifyIndex);
+            if (AgentVersion >= SemanticVersion.Parse("1.7.0")) {
+                Assert.NotEqual(0, root.CreateIndex);
+                Assert.NotEqual(0, root.ModifyIndex);
+                Assert.NotEqual(0, root.SerialNumber);
+            }
         }
 
         [SkippableFact]
--- a/Consul.Test/AgentTest.cs
+++ b/Consul.Test/AgentTest.cs
@@ -1044,9 +1044,9 @@ namespace Consul.Test
             Assert.Null(root.IntermediateCerts);
             Assert.True(root.Active);
             Assert.NotNull(root.PrivateKeyType);
-            Assert.NotEqual(0, root.PrivateKeyBits);
             if (AgentVersion >= SemanticVersion.Parse("1.7.0"))
             {
+                Assert.NotEqual(0, root.PrivateKeyBits);
                 Assert.NotEqual(0, root.CreateIndex);
                 Assert.NotEqual(0, root.ModifyIndex);
                 Assert.NotEqual(0, root.SerialNumber);
--- a/Consul.Test/AgentTest.cs
+++ b/Consul.Test/AgentTest.cs
@@ -1030,8 +1030,24 @@ namespace Consul.Test
         {
             var caRoots = await _client.Agent.GetCARoots();
             Assert.NotEqual((ulong)0, caRoots.LastIndex);
-            Assert.Single(caRoots.Response.Roots);
+            Assert.NotNull(caRoots.Response.ActiveRootID);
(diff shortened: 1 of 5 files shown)


IvanKolchanov:
The only issue I have is with another test case that the Consul in Golang has: https://github.com/hashicorp/consul/blob/v1.9.17/api/connect_ca_test.go#L12 .
They use the testutil to create another client for testing and do not bootstrap connect in that test case, which is, however, not really possible with the current testing setup that consuldotnet has. Because that would require to create another test_config.json and run both the new and the old one when testing. Please correct me if I am wrong!

marcin-krystianc:
> The only issue I have is with another test case that the Consul in Golang has: https://github.com/hashicorp/consul/blob/v1.9.17/api/connect_ca_test.go#L12 . They use the testutil to create another client for testing and do not bootstrap connect in that test case, which is, however, not really possible with the current testing setup that consuldotnet has. Because that would require to create another test_config.json and run both the new and the old one when testing. Please correct me if I am wrong!

You are right, but we don't have to replicate that test. The test you've already added is sufficient. It tests that the endpoint works and returns desired data.

Consul.Test/AgentTest.cs
```diff
@@ -1025,6 +1025,15 @@ public async Task Agent_Metrics()
             Assert.NotNull(agentMetrics.Response.Samples);
         }
 
+        [Fact]
+        public async Task Agent_CARoots()
+        {
+            var caRoots = await _client.Agent.GetCARoots();
+            Assert.NotEqual((ulong)0, caRoots.LastIndex);
+            Assert.Single(caRoots.Response.Roots);
+            Assert.Equal("11111111-2222-3333-4444-555555555555.consul", caRoots.Response.TrustDomain);
```
marcin-krystianc, line 1034: This test could test all fields in the response (at least for fields being non-null). That would assure us that all fields in the newly added structures are named correctly.

marcin-krystianc (commented):
Generally, it looks good. Please add more test assertions so we know that all the fields are named correctly.

IvanKolchanov:
> > The only issue I have is with another test case that the Consul in Golang has: https://github.com/hashicorp/consul/blob/v1.9.17/api/connect_ca_test.go#L12 . They use the testutil to create another client for testing and do not bootstrap connect in that test case, which is, however, not really possible with the current testing setup that consuldotnet has. Because that would require to create another test_config.json and run both the new and the old one when testing. Please correct me if I am wrong!
> 
> You are right, but we don't have to replicate that test. The test you've already added is sufficient. It tests that the endpoint works and returns desired data.

That's what I thought, just had to make sure! :)

Consul.Test/AgentTest.cs
IvanKolchanov: Just added the test assertions, ready for further review!

IvanKolchanov:
@marcin-krystianc I see that a few tests had failed because of an assertion error, expected that a bit, because I wasn't sure which kind of values can SerialNumber, CreateIndex and ModifyIndex take depending on conditions. PrivateKeyBits shouldn't be 0 though. Don't think that there is a way for me to initialize those values in config, except for PrivateKeyBits (https://developer.hashicorp.com/consul/docs/connect/ca/consul). Should I just delete the assert tests for SerialNumber, CreateIndex and ModifyIndex?
Thank you in advance!

marcin-krystianc:
> @marcin-krystianc I see that a few tests had failed because of an assertion error, expected that a bit, because I wasn't sure which kind of values can SerialNumber, CreateIndex and ModifyIndex take depending on conditions. PrivateKeyBits shouldn't be 0 though. Don't think that there is a way for me to initialize those values in config, except for PrivateKeyBits (https://developer.hashicorp.com/consul/docs/connect/ca/consul). Should I just delete the assert tests for SerialNumber, CreateIndex and ModifyIndex? Thank you in advance!

It seem that the problem only occurs for version 1.6.10. So perhaps we can use `AgentVersion` variable (https://github.com/G-Research/consuldotnet/blob/219ecd42b53d7b0114aea540e5fb9a0fbce1a56e/Consul.Test/BaseFixture.cs#L15) and assert those problematic fields only when the version is equal or greater than `1.7.0` ?

Consul.Test/AgentTest.cs
```diff
@@ -1025,6 +1025,34 @@ public async Task Agent_Metrics()
...
+            Assert.NotNull(root.Name);
+            Assert.NotNull(root.SigningKeyID);
+            Assert.NotNull(root.ExternalTrustDomain);
+            Assert.NotNull(root.NotBefore);
+            Assert.NotNull(root.NotAfter);
+            Assert.NotNull(root.RootCert);
+            Assert.Null(root.IntermediateCerts);
+            Assert.True(root.Active);
+            Assert.NotNull(root.PrivateKeyType);
+            Assert.NotEqual(0, root.PrivateKeyBits);
```
marcin-krystianc, line 1047: I think this one needs to go inside the `if (AgentVersion >= SemanticVersion.Parse("1.7.0"))` block.

IvanKolchanov: Yeah, a bit surprising, but should be working now, fixed it!

marcin-krystianc (approved):
I think now it is perfect, thank you!

PR #338, opened 2024-06-19
Added <summary> to GetCARoots, added GetCARoots with QueryOptions parameter

IvanKolchanov:


IvanKolchanov:
@marcin-krystianc the PR with fixes to #332

Consul/Agent.cs
```diff
@@ -1164,11 +1164,27 @@ public async Task<QueryResult<ServiceConfiguration>> GetServiceConfiguration(str
...
         }
 
+        /// <summary>
+        /// GetCARoots returns root certificates in the cluster
+        /// </summary>
+        /// <param name="ct">Cancellation Token</param>
+        /// <returns>Root certificates</returns>
         public async Task<QueryResult<CARoots>> GetCARoots(CancellationToken ct = default)
         {
             return await _client.Get<CARoots>("v1/agent/connect/ca/roots", QueryOptions.Default).Execute(ct).ConfigureAwait(false);
```
marcin-krystianc, line 1174: Can you call the 
```suggestion
            return await GetCARoots(QueryOptions.Default, ct);
```

IvanKolchanov: Oh, makes sense. Shouldn't it be "return await GetCARoots(QueryOptions.Default, ct).ConfigureAwait(false);" though?

marcin-krystianc: yeah, you are right.

marcin-krystianc (approved):
Thanks!

PR #340, opened 2024-06-20
[CI] Enable manual CI runs

marcin-krystianc:
We have a "formatter" workflow that applies automatic code changes. The downside of applying automatic code changes is that they do not trigger other workflows (e.g., the "CI" workflow). To work around this problem, we can enable manual workflow runs.


jgiannuzzi (approved):
LGTM but we should find a real fix for this in the long run

PR #348, opened 2024-07-15
Added GET Update LAN Coordinates endpoints, added tests for it

### LINKED ISSUE #322: FEATURE REQUEST: Update LAN Coordinates for a node
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Add support to [Update LAN Coordinates for a node](https://developer.hashicorp.com/consul/api-docs/v1.9.x/coordinate#update-lan-coordinates-for-a-node) 
This endpoint updates the LAN network coordinates for a node in a given datacenter.
This API is supported in the Golang client ([1.9.17](https://github.com/hashicorp/consul/blob/4897b16abf192942d1e1d2ffa109bf17b2431e9a/api/coordinate.go#L71))

### Describe alternatives you've considered

Nothing for the moment.

### Additional context

Nothing for the moment.

IvanKolchanov:
Resolves #322 
@marcin-krystianc

### DIFF
--- a/Consul.Test/CoordinateTest.cs
+++ b/Consul.Test/CoordinateTest.cs
@@ -96,7 +96,7 @@ namespace Consul.Test
             for (int i = 0; i < 8; i++) coord.Coord.Vec.Add(0.0);
 
             var response = await _client.Coordinate.Update(coord);
-            Assert.Equal("OK", response.StatusCode.ToString());
+           Assert.Equal(HttpStatusCode.OK, response.StatusCode);;
 
             var newCoordResult = await _client.Coordinate.Node(nodeName);
             for (int i = 0; i < 5; i++)
--- a/Consul.Test/CoordinateTest.cs
+++ b/Consul.Test/CoordinateTest.cs
@@ -73,5 +73,46 @@ namespace Consul.Test
             Assert.IsType<CoordinateEntry[]>(nodeDetails);
             Assert.NotEmpty(nodeDetails);
         }
+
+        [Fact]
+        public async Task Coordinate_Update()
+        {
+            var nodeName = "foo-update-lan";
+            var registration = new CatalogRegistration
+            {
+                Node = nodeName,
+                Address = "1.1.1.1"
+            };
+            var register = await _client.Catalog.Register(registration);
+
+            var coord = new CoordinateEntry
+            {
+                Node = nodeName,
+                Coord = new SerfCoordinate()
+            };
+            coord.Coord.Error = 1.5;
+            coord.Coord.Height = 0.5;
+            coord.Coord.Adjustment = 0.0;
+            for (int i = 0; i < 8; i++) coord.Coord.Vec.Add(0.0);
+            
+            var response = await _client.Coordinate.Update(coord);
+            Assert.Equal("OK", response.StatusCode.ToString());
+
+            var newCoordResult = await _client.Coordinate.Node(nodeName);
+            for (int i = 0; i < 4; i++)
+            {
+                if (newCoordResult != null) break;
+                await Task.Delay(1000);
+                newCoordResult = await _client.Coordinate.Node(nodeName);
+            }
+
+            Assert.NotNull(newCoordResult);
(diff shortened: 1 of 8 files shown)


Consul.Test/CoordinateTest.cs
```diff
@@ -73,5 +73,46 @@ public async Task Coordinate_GetNode()
...
+
+            var coord = new CoordinateEntry
+            {
+                Node = nodeName,
+                Coord = new SerfCoordinate()
+            };
+            coord.Coord.Error = 1.5;
+            coord.Coord.Height = 0.5;
+            coord.Coord.Adjustment = 0.0;
+            for (int i = 0; i < 8; i++) coord.Coord.Vec.Add(0.0);
```
marcin-krystianc, line 96:
```suggestion
            var coord = new CoordinateEntry
            {
                Node = nodeName,
                Coord = new SerfCoordinate
                {
                    Error = 1.5,
                    Height = 0.5,
                    Adjustment = 0.0,
                    Vec = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0},
                }
            };
```

Consul.Test/CoordinateTest.cs
```diff
@@ -73,5 +73,46 @@ public async Task Coordinate_GetNode()
...
+
+            var response = await _client.Coordinate.Update(coord);
+            Assert.Equal("OK", response.StatusCode.ToString());
+
+            var newCoordResult = await _client.Coordinate.Node(nodeName);
+            for (int i = 0; i < 5; i++)
```
marcin-krystianc, line 102: I'm not a fan of that loop.
I think there is a reliable way to query the node without using the loop.
You can use so called "blocking queries", i.e.: use query options with a non-zero wait-index:

```
   await _client.Catalog.Register(registration);
   var nodeResult = await _client.Catalog.Node(nodeName);
   Assert.Equal(HttpStatusCode.OK, nodeResult.StatusCode);
...
   var q = new QueryOptions{ WaitIndex = nodeResult.LastIndex, };
   var newCoordResult = await _client.Coordinate.Node(nodeName, q);
   Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```

I tried this locally and it seemed working reliably.

marcin-krystianc (commented):
Left some suggestions

Consul.Test/CoordinateTest.cs
IvanKolchanov: That's fair, there wasn't really a similar situation in the previous code, so I went for something similar to Golang code's retry loop. Should we change waitTime to 10 seconds, to avoid hanging for the default 5 minutes? And I have a little question, I am a bit confused on how is the index of node registered in catalog connected to the coordinate node? I am still working on my familiarity with all consul features, sorry.

marcin-krystianc: No worries, it is sometimes a bit of unknown for me as well 😄 I'm concerned about the loop because it introduces a risk of flakiness. We run each test many times (we have a matrix with os, consul version and runtime version), thus we need to have really robust tests.
I think the "LastIndex"/"WaitIndex" are global, so it doesn't matter that they are from different endpoints. 

> Should we change waitTime to around 10 seconds, to avoid hanging for the default 5 minutes
I think it is better to wait longer, but have more robust tests. Therefore I wouldn't change the default waitTime.

IvanKolchanov: Thanks!) Makes sense, I will commit all the changes and we see how the tests go.

marcin-krystianc: FYI: I've just updated my code snipped from above, to make it clear that the `Catalog.Node()` call should be before the `Coordinate.Update()` call.

IvanKolchanov:
@marcin-krystianc 

> Left some suggestions

Okey, everything should be good

marcin-krystianc (dismissed):
Nice 👍

marcin-krystianc (approved):

Issue #373, opened 2024-10-14
[FEATURE REQUEST]:  Add support for List Network Segments


larrytamnjong:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Implement support for the [List Network Segments Endpoint ](https://developer.hashicorp.com/consul/api-docs/v1.11.x/operator/segment) present in consul client 1.11.11

PR #374, opened 2024-10-14
add support for list segments endpoint

larrytamnjong:
Added support for List segments Endpoint #373

### DIFF
--- a/Consul.Test/OperatorTest.cs
+++ b/Consul.Test/OperatorTest.cs
@@ -94,7 +94,7 @@ namespace Consul.Test
         public async Task Segment_List()
         {
             var segments = await _client.Operator.SegmentList();
-            Assert.IsType<string[]>(segments.Response);
+            Assert.NotEmpty(segments.Response);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


larrytamnjong:
@marcin-krystianc

Consul.Test/OperatorTest.cs
```diff
@@ -89,5 +89,12 @@ public async Task Operator_GetLicense()
             var queryResult = await _client.Operator.GetConsulLicense();
             Assert.NotNull(queryResult.Response);
         }
+
+        [EnterpriseOnlyFact]
+        public async Task Segment_List()
+        {
+            var segments = await _client.Operator.SegmentList();
+            Assert.IsType<string[]>(segments.Response);
```
marcin-krystianc, line 97: Maybe we could check that the response is not empty (Check if the length of the array is non-zero?)

larrytamnjong: Thanks @marcin-krystianc I have made the update

marcin-krystianc (approved):
👍

Issue #379, opened 2024-10-19
[FEATURE REQUEST]:  Add Support for Admin Partition - API


larrytamnjong:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Implement support for the [Admin Partition  API ](https://developer.hashicorp.com/consul/api-docs/v1.11.x/admin-partitions) found in v1.11.11 of the goland library

PR #380, opened 2024-10-20
add admin partition create API support (#379)

larrytamnjong:


larrytamnjong:
@marcin-krystianc

marcin-krystianc (commented):
These changes look good, but because we are unable to test it I think we should park these PR for now. If it will be possible to run enterprise tests for versions newer than `1.10`, we will get back to it.

PR #382, opened 2024-10-21
Operator Create Network Area

larrytamnjong:
@marcin-krystianc

### DIFF
--- a/Consul/Interfaces/IOperatorEndpoint.cs
+++ b/Consul/Interfaces/IOperatorEndpoint.cs
@@ -45,8 +45,8 @@ namespace Consul
         Task<QueryResult<ConsulLicense>> GetConsulLicense(string datacenter = "", CancellationToken ct = default);
         Task<QueryResult<string[]>> SegmentList(QueryOptions q, CancellationToken cancellationToken = default);
         Task<QueryResult<string[]>> SegmentList(CancellationToken cancellationToken = default);
-        Task<WriteResult<string>> CreateArea(Area area, WriteOptions q, CancellationToken ct = default);
-        Task<WriteResult<string>> CreateArea(Area area, CancellationToken ct = default);
+        Task<WriteResult<AreaResponse>> CreateArea(Area area, WriteOptions q, CancellationToken ct = default);
+        Task<WriteResult<AreaResponse>> CreateArea(Area area, CancellationToken ct = default);
 
     }
 }
--- a/Consul/Interfaces/IOperatorEndpoint.cs
+++ b/Consul/Interfaces/IOperatorEndpoint.cs
@@ -45,6 +45,8 @@ namespace Consul
         Task<QueryResult<ConsulLicense>> GetConsulLicense(string datacenter = "", CancellationToken ct = default);
         Task<QueryResult<string[]>> SegmentList(QueryOptions q, CancellationToken cancellationToken = default);
         Task<QueryResult<string[]>> SegmentList(CancellationToken cancellationToken = default);
+        Task<WriteResult<string>> CreateArea(Area area, WriteOptions q, CancellationToken ct = default);
+        Task<WriteResult<string>> CreateArea(Area area, CancellationToken ct = default);
 
     }
 }
--- a/Consul/Interfaces/IOperatorEndpoint.cs
+++ b/Consul/Interfaces/IOperatorEndpoint.cs
@@ -45,8 +45,8 @@ namespace Consul
         Task<QueryResult<ConsulLicense>> GetConsulLicense(string datacenter = "", CancellationToken ct = default);
         Task<QueryResult<string[]>> SegmentList(QueryOptions q, CancellationToken cancellationToken = default);
         Task<QueryResult<string[]>> SegmentList(CancellationToken cancellationToken = default);
-        Task<WriteResult<AreaResponse>> CreateArea(Area area, WriteOptions q, CancellationToken ct = default);
-        Task<WriteResult<AreaResponse>> CreateArea(Area area, CancellationToken ct = default);
+        Task<WriteResult<string>> CreateArea(AreaRequest area, WriteOptions q, CancellationToken ct = default);
+        Task<WriteResult<string>> CreateArea(AreaRequest area, CancellationToken ct = default);
 
     }
 }
--- a/Consul/Operator.cs
+++ b/Consul/Operator.cs
@@ -100,7 +100,7 @@ namespace Consul
         public int NumNodes { get; set; }
     }
 
-    public class Area
+    public class AreaRequest
     {
         /// <summary>
         /// PeerDatacenter is the peer Consul datacenter that will make up the
@@ -123,7 +123,7 @@ namespace Consul
         public bool UseTLS { get; set; }
     }
 
-    public class AreaResponse : Area
+    public class Area : AreaRequest
     {
         /// <summary>
         /// ID is this identifier for an area (a UUID). This must be left empty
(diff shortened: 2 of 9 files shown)


Consul/Operator.cs
```diff
@@ -100,6 +100,34 @@ public class KeyringResponse
         public int NumNodes { get; set; }
     }
 
+    public class Area
```
marcin-krystianc, line 103: Please split it into two structures, one for request and one for response.

Consul/Operator.cs
```diff
@@ -100,6 +100,37 @@ public class KeyringResponse
...
+        public string[] RetryJoin { get; set; }
+
+        /// <summary>
+        /// UseTLS specifies whether gossip over this area should be encrypted with TLS
+        /// if possible.
+        /// </summary>
+        public bool UseTLS { get; set; }
+    }
+
+    public class AreaResponse : Area
```
marcin-krystianc, line 126: Inheritance is not needed here, please remove it.

Consul/Interfaces/IOperatorEndpoint.cs
```diff
@@ -45,6 +45,8 @@ public interface IOperatorEndpoint
         Task<QueryResult<ConsulLicense>> GetConsulLicense(string datacenter = "", CancellationToken ct = default);
         Task<QueryResult<string[]>> SegmentList(QueryOptions q, CancellationToken cancellationToken = default);
         Task<QueryResult<string[]>> SegmentList(CancellationToken cancellationToken = default);
+        Task<WriteResult<string>> CreateArea(Area area, WriteOptions q, CancellationToken ct = default);
```
marcin-krystianc, line 48: The result type should be `Task<WriteResult<AreaResponse>>`

Consul/Operator.cs
larrytamnjong: Thanks, @marcin-krystianc  Initially, I considered using inheritance because, in certain cases like the[ List Specific Network Area ](https://developer.hashicorp.com/consul/api-docs/v1.11.x/operator/area#list-specific-network-area) response, additional properties such as PeerDatacenter and RetryJoin are returned. I thought that inheriting from the Area class would make the response more flexible for these scenarios.

marcin-krystianc: > Thanks, @marcin-krystianc Initially, I considered using inheritance because, in certain cases like the[ List Specific Network Area ](https://developer.hashicorp.com/consul/api-docs/v1.11.x/operator/area#list-specific-network-area) response, additional properties such as PeerDatacenter and RetryJoin are returned. I think inheriting from the Area class will make the response more flexible for these scenarios.

Ah, I see. It is tempting to reuse existing structure definitions for new ones to save some code, but in my opinion, it isn't always the right thing to do. Given that apart from `Create Network Area` there are other methods like `List Network Areas`, `Update Network Area`, and others - we need to plan how we are going to define requests and responses for all of them. 
My initial plan was to create a dedicated request and response structure definition for each method. That would give us full control over those structures and would define the interface very clearly from a user perspective (It is more code in the library to write but it is very straightforward).
On the other hand, I looked at the consul implementation of Operator_Area API (https://github.com/hashicorp/consul/blob/main/api/operator_area.go) and maybe it is ok to do it similarly.
It would require fewer structure definitions but I guess the API will be less clear from the user perspective (for example `AreaUpdate` methods take `ID` and `Area` objects, but `Area` has already an `ID` but it is not used).

I think I'll leave it to you to decide. Both approaches have pros and cons

larrytamnjong: Thanks for the insights @marcin-krystianc ! I initially removed the inheritance and made adjustments based on your initial recommendations, but since you're leaving it up to me, I've added it back. Please let me know what you think!

marcin-krystianc: The use of inheritance somewhat bothers me, but maybe unnecessarily :-) The only suggestion I have is to change the names: `Area` -> `AreaRequest`  and `AreaResponse` -> `Area`

Consul/Operator.cs
```diff
@@ -249,6 +280,25 @@ public Task<QueryResult<string[]>> SegmentList(CancellationToken ct = default)
...
+        }
+
+        /// <summary>
+        /// CreateArea will create a new network area. The ID in the given structure must
+        /// be empty and a generated ID will be returned on success.
+        /// </summary>
+        public async Task<WriteResult<AreaResponse>> CreateArea(Area area, WriteOptions q, CancellationToken ct = default)
+        {
+            var req = await _client.Post<Area, AreaResponse>("/v1/operator/area", area, q).Execute(ct).ConfigureAwait(false);
+            return new WriteResult<AreaResponse>(req, req.Response);
```
marcin-krystianc, line 300: Returning an object, where only the `ID` field will be initialised, but others will be not - is not great from the usability perspective. (It is confusing, to se a struct which is not fully initialised, you need an extra effort to refer to docs or to implementation to understand why is that)
In the go implementation (https://github.com/hashicorp/consul/blob/main/api/operator_area.go#L91), the `Create` method returns string, so I guess that what we also should do(i.e. `return new WriteResult<string>(req, req.Response.ID)`).

Initially I wanted a dedicated struct for response, because I anticipated other fields being added to the response. But since go implementation returns only a single string, we can do it too.

Consul/Interfaces/IOperatorEndpoint.cs
marcin-krystianc: > The result type should be `Task<WriteResult<AreaResponse>>`

When I made that comment I assumed that `AreaResponse` will contain only an `ID` field.

Consul/Operator.cs
larrytamnjong: Thank you so much, @marcin-krystianc for the detailed explanation and guidance! 😊 I've made the changes you suggested. Please let me know if there’s anything else I should adjust.

Consul/Operator.cs
```diff
@@ -100,6 +100,37 @@ public class KeyringResponse
...
+        /// UseTLS specifies whether gossip over this area should be encrypted with TLS
+        /// if possible.
+        /// </summary>
+        public bool UseTLS { get; set; }
+    }
+
+    public class Area : AreaRequest
+    {
+        /// <summary>
+        /// ID is this identifier for an area (a UUID). This must be left empty
```
marcin-krystianc, line 129: With our API, the comment `This must be left empty when creating a new area.` is not relevant.

Consul/Operator.cs
```diff
@@ -249,6 +280,25 @@ public Task<QueryResult<string[]>> SegmentList(CancellationToken ct = default)
         {
             return SegmentList(QueryOptions.Default, ct);
         }
+
+        /// <summary>
+        /// CreateArea will create a new network area. The ID in the given structure must
```
marcin-krystianc, line 285: here as well -> ` The ID in the given structure must be empty and a generated ID will be returned on success.`

larrytamnjong: Done, thanks.

Consul/Operator.cs
```diff
@@ -249,6 +279,23 @@ public Task<QueryResult<string[]>> SegmentList(CancellationToken ct = default)
         {
             return SegmentList(QueryOptions.Default, ct);
         }
+
+        /// <summary>
+        /// CreateArea will create a new network area.
```
marcin-krystianc, line 284: I think we should leave the part with returned ID though:
```suggestion
        /// CreateArea will create a new network area, a generated ID will be returned on success.
```

Consul/Operator.cs
```diff
@@ -249,6 +279,23 @@ public Task<QueryResult<string[]>> SegmentList(CancellationToken ct = default)
...
+        /// </summary>
+        public Task<WriteResult<string>> CreateArea(AreaRequest area, CancellationToken ct = default)
+        {
+            return CreateArea(area, WriteOptions.Default, ct);
+        }
+
+        /// <summary>
+        /// CreateArea will create a new network area.
```
marcin-krystianc, line 292:
```suggestion
        /// CreateArea will create a new network area, a generated ID will be returned on success.
```

Consul/Operator.cs
larrytamnjong: I deleted that without thinking 😂—thanks for pointing it out!

marcin-krystianc (approved):
Looks good now, thanks for bearing with me.

Issue #381, opened 2024-10-21
[FEATURE REQUEST]: Add support for Network Areas - Operator HTTP API


larrytamnjong:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Implement support for the [Network Areas - Operator HTTP API](https://developer.hashicorp.com/consul/api-docs/v1.11.x/operator/area) found in v1.11.11 of the goland library

PR #389, opened 2024-11-01
add area update support and test (#381)

larrytamnjong:


larrytamnjong:
@marcin-krystianc

Consul.Test/OperatorTest.cs
```diff
@@ -123,5 +123,17 @@ public async Task Operator_AreaList()
...
+        public async Task Operator_AreaUpdate()
+        {
+            var peerDataCenter = KVTest.GenerateTestKeyName();
+            var area = new AreaRequest { PeerDatacenter = peerDataCenter, UseTLS = false, RetryJoin = null };
+            var createResponse = await _client.Operator.AreaCreate(area);
+
+            area = new AreaRequest { PeerDatacenter = peerDataCenter, UseTLS = true, RetryJoin = new string[] { "10.1.2.9", "10.1.2.0" } };
+            var updateResponse = await _client.Operator.AreaUpdate(area, createResponse.Response);
+
+            Assert.NotNull(updateResponse.Response);
```
marcin-krystianc, line 136: Can we check what is exactly the content in the response?
Also it would be great to check if the Area properties,  before and after the update, are different.

marcin-krystianc (commented):
The code is correct, but I would like to encourage you to think more about code readability.
Using right variable names can make huge difference for the mental effort needed to digest the code.

Consul.Test/OperatorTest.cs
```diff
@@ -123,5 +123,24 @@ public async Task Operator_AreaList()
...
+            var area = new AreaRequest { PeerDatacenter = peerDataCenter, UseTLS = false, RetryJoin = null };
+            var createResult = await _client.Operator.AreaCreate(area);
+            var areaId = createResult.Response;
+
+            area = new AreaRequest { PeerDatacenter = peerDataCenter, UseTLS = true, RetryJoin = new string[] { "10.1.2.9", "10.1.2.0" } };
+            var updateResult = await _client.Operator.AreaUpdate(area, createResult.Response);
+
+            var listResult = await _client.Operator.AreaList();
+            var updatedArea = listResult.Response.Single(x => x.ID == areaId);
+
```
marcin-krystianc, line 139: nit: Extra empty line

Consul.Test/OperatorTest.cs
```diff
@@ -123,5 +123,24 @@ public async Task Operator_AreaList()
...
+        [EnterpriseOnlyFact]
+        public async Task Operator_AreaUpdate()
+        {
+            var peerDataCenter = KVTest.GenerateTestKeyName();
+            var area = new AreaRequest { PeerDatacenter = peerDataCenter, UseTLS = false, RetryJoin = null };
+            var createResult = await _client.Operator.AreaCreate(area);
+            var areaId = createResult.Response;
+
+            area = new AreaRequest { PeerDatacenter = peerDataCenter, UseTLS = true, RetryJoin = new string[] { "10.1.2.9", "10.1.2.0" } };
+            var updateResult = await _client.Operator.AreaUpdate(area, createResult.Response);
```
marcin-krystianc, line 135:
```suggestion
            var updateResult = await _client.Operator.AreaUpdate(area, areaId);
```

marcin-krystianc: Assertion for `updateResult` is missing now, please add it back 😄

larrytamnjong: I removed it because we're already using the initial createResponse areaId here: ```var updatedArea = listResult.Response.Single(x => x.ID == areaId);``` to fetch the updated query. This somehow asserts that createResult areaId is equal to updatedResult areaId, so the additional assertion is seemed redundant. I guess I was over thinking it. I have added it back now—thanks 😊

marcin-krystianc: Thanks, one last thing. Please apply the suggested change for using `areaId` instead of `createResult.Response` (If I do it I'll not be able to approve the PR).

marcin-krystianc (approved):

PR #392, opened 2024-11-06
add area get support and test (#381)

larrytamnjong:


larrytamnjong:
@marcin-krystianc

marcin-krystianc (approved):
It is a little bit weird that the get method returns an array, but yeah that it is what the underlying API call does 🤷

PR #397, opened 2024-11-19
add area join support and test (#381)

larrytamnjong:


Consul.Test/OperatorTest.cs
larrytamnjong, line 1: Hi @marcin-krystianc
I'm confused about this. Which address should I provide for this test?

marcin-krystianc: Did you try to run it locally ? What are the logs in the consul agent?

larrytamnjong: It's an EnterpriseOnlyFact so it skips when running locally.

marcin-krystianc: It is possible to run enterprise tests locally. You can download the old version (e.g. 1.9+ent of enterprise consul which will run locally for an hour). Then you need either set the appropriate env variable or just replace the tests attribute with `[Fact]`.

larrytamnjong: @marcin-krystianc here is the response from the consul agent 
image

marcin-krystianc: I don't know that consul feature, so I'm not sure if it is even possible to make a successful area join request. 
In worst case we can have a test with an unsuccessful call and an assertion for the exception we get back.

larrytamnjong: Thanks @marcin-krystianc I guess i can mark this as a draft and come back to it later.

marcin-krystianc:
It hasn't been updated for 2 months, so I'm closing for now. Feel free to reopen.

PR #407, opened 2024-12-09
Making sure that we are not susceptible to user supplied executables

marcin-krystianc:


dgl (approved):
Seems fine; assuming `dotnet-format` is actually on the PATH...

PR #412, opened 2025-01-03
Embed PDBs

sungam3r:
Replaces #14

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,7 +10,6 @@
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <DebugType>embedded</DebugType>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <Authors>G-Research</Authors>
     <PublicSign>true</PublicSign>
     <AssemblyOriginatorKeyFile>../assets/consuldotnet.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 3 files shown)


Directory.Build.props
```diff
@@ -8,8 +8,9 @@
     <PackageReleaseNotes>https://github.com/G-Research/consuldotnet/blob/master/CHANGELOG.md</PackageReleaseNotes>
     <PackageProjectUrl>https://consuldot.net</PackageProjectUrl>
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
-    <RepositoryType>git</RepositoryType>
-    <RepositoryUrl>https://github.com/G-Research/consuldotnet</RepositoryUrl>
+    <DebugType>embedded</DebugType>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
```
marcin-krystianc, line 13: I don't think this is necessary, as I I expect all sources to be tracked by git.

sungam3r: OK. We can check it after publish.

sungam3r:
```suggestion
```

Directory.Build.props
```diff
@@ -8,8 +8,8 @@
     <PackageReleaseNotes>https://github.com/G-Research/consuldotnet/blob/master/CHANGELOG.md</PackageReleaseNotes>
     <PackageProjectUrl>https://consuldot.net</PackageProjectUrl>
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
-    <RepositoryType>git</RepositoryType>
-    <RepositoryUrl>https://github.com/G-Research/consuldotnet</RepositoryUrl>
```
marcin-krystianc, line 12: I was wondering why we don't need to add the reference to the `Microsoft.SourceLink.GitHub` package, but according to the docs (https://github.com/dotnet/sourcelink?tab=readme-ov-file#using-source-link-in-net-projects), it is not needed anymore (starting with .NET 8).

marcin-krystianc (approved):
Thanks [sungam3r](https://github.com/sungam3r), it looks good 👍

Directory.Build.props
sungam3r: yep

PR #413, opened 2025-01-03
Configure JSON serializer

sungam3r:
Small change to allow configuration of used JSON serializer, no breaking changes. I know about #43 and #158 but it seems work is stalled. I suggest to do a small step forward on existing serializer to allow configure it. I was faced with the fact that I can’t configure the serialization of some request metadata as it is necessary.

### DIFF
--- a/Consul/Client.cs
+++ b/Consul/Client.cs
@@ -23,6 +23,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Cryptography.X509Certificates;
 using Consul.Filtering;
+using Newtonsoft.Json;
 
 namespace Consul
 {
@@ -61,6 +62,11 @@ namespace Consul
                 OnUpdated(new EventArgs());
             }
         }
+
+        /// <summary>
+        /// Allows to configure the serialization settings for a Consul client.
+        /// </summary>
+        public Action<JsonSerializerSettings> ConfigureSerializerSettings { get; set; }
 
         /// <summary>
         /// The Uri to connect to the Consul agent.
--- a/Consul/Client_Request.cs
+++ b/Consul/Client_Request.cs
@@ -57,7 +57,7 @@ namespace Consul
         internal Stream ResponseStream { get; set; }
         internal string Endpoint { get; set; }
 
-        internal readonly JsonSerializer _serializer = new JsonSerializer();
+        internal readonly JsonSerializer _serializer;
 
         internal ConsulRequest(ConsulClient client, string url, HttpMethod method)
         {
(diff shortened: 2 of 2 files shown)


marcin-krystianc:
Hi @sungam3r,

can you give more details about your use case? Ideally, we want to avoid adding a public method with an argument type from `Newtonsoft.Json` (I'm asking, because it makes it harder to replace `Netwtonsoft.Json` with `System.Text.Json` in the future).

sungam3r:
I understand your reasoning about public method referencing nsj, but migration to stj will be a breaking change anyway - other lib dependency, other attributes, other converters, etc. Besides, PR with this migration looks abandoned and honestly I do not feel that this migration happens in the near future.

Now about my case. I'm not going to change some default settings here but rather want to **revert** them to be default ones because some component in my process environment hijack default serializer settings globally and so it badly influences on serialization performed by consul client. I just want to be able to specify serialization settings particularly for consul client to make it stable and independent from global settings.

sungam3r:
@marcin-krystianc Do you need additional information from me?

marcin-krystianc:
> @marcin-krystianc Do you need additional information from me?

Hi [sungam3r](https://github.com/sungam3r), I'm sorry I was occupied last week. I'm just now starting to work on this issue.

marcin-krystianc:
> Now about my case. I'm not going to change some default settings here but rather want to revert them to be default ones because some component in my process environment hijack default serializer settings globally and so it badly influences on serialization performed by consul client. I just want to be able to specify serialization settings particularly for consul client to make it stable and independent from global settings.

Could you share a code snipped that will let me reproduce the issue?

sungam3r:
Or I can push commit trying to do the same myself. Ok?

marcin-krystianc:
> I have no snippet but I'm sure you may reproduce behavior simply adding something like this
> 
> ```
> JsonConvert.DefaultSettings = () => new JsonSerializerSettings
> {
>     Formatting = some_unusual_value,
>     TypeNameHandling = some_unusual_value,
>     ContractResolver = new some_unusual_contract_resolver_printing_awkward_things
> };
> ```
> 
> in the beginning of test initialization.

Thanks, indeed it is very easy to break our library by overriding the default serializer settings.
I am still not convinced that exposing Newtonsoft even more in the public API is the right thing to do. Shouldn't we try to make the library independent from overriding the DefaultSettings? That wouldn't change the public API, but it would help with your problem.

marcin-krystianc:
@sungam3r 
I opened an alternative PR. It fixes the issue but doesn't require API changes -> https://github.com/G-Research/consuldotnet/pull/417

sungam3r:
I'm fine with such alternative. It suits my needs. Thanks.

sungam3r:
Replaced by #417 .

PR #416, opened 2025-01-21
[CI] Update .NET versions and reduce test matrix cardinality

marcin-krystianc:
- `.NET 8` is the latest LTS
- `.NET 9` is the latest STS (Standard Term Support)
- `.NET 6` and `.NET 7` are out of support
- taking out the `macos-13` from the matrix and adding a single job with it to reduce matrix cardinality

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,36 +58,33 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


adamreeve (approved):
> taking out the macos-13 from the matrix and adding a single job with it to reduce matrix cardinality

Should we do the same thing for `Consul_AspNetCore` for consistency?

marcin-krystianc:
> > taking out the macos-13 from the matrix and adding a single job with it to reduce matrix cardinality
> 
> Should we do the same thing for `Consul_AspNetCore` for consistency?

Probably not needed, For `Consul_AspNetCore` we don't have massive cardinality because we don't run tests for every consul version. I think it is good as it is now.

PR #419, opened 2025-01-21
[CI] Fixing tests when running locally

marcin-krystianc:


Consul.Test/CatalogTest.cs
```diff
@@ -168,7 +169,7 @@ public async Task Catalog_GetTaggedAddressesService()
 
             await _client.Catalog.Register(registration);
 
-            var services = await _client.Catalog.Service("redis");
+            var services = await _client.Catalog.Service(svcID2);
```
marcin-krystianc, line 172: We need to use a unique name to avoid any ambiguity.

Consul.Test/DiscoveryChainTest.cs
```diff
@@ -80,7 +80,7 @@ public async Task DiscoveryChain_Get()
             Assert.Equal(check.ServiceName, chain.ServiceName);
             Assert.Equal(check.Namespace, chain.Namespace);
             Assert.Equal(check.Datacenter, chain.Datacenter);
-            Assert.Equal(check.Protocol, chain.Protocol);
+            Assert.NotEmpty(chain.Protocol);
```
marcin-krystianc, line 83: Protocol can be either `tcp` or `http`.

PR #427, opened 2025-02-20
ListIntentions

FortunateOmonuwa:
**This PR implements retrieving a list of service intentions and updates related classes.**

**Changes Introduced:**
- Implements fetching a list of intentions.
- Updates ServiceIntentionsEntry by adding:
  - List<SourceIntention> Sources.

- Adds new classes:
  - SourceIntention
  - IntentionPermission
  - IntentionHTTPPermission
  - IntentionHTTPHeaderPermission

### DIFF
--- a/Consul.Test/ConfigurationTest.cs
+++ b/Consul.Test/ConfigurationTest.cs
@@ -22,9 +22,9 @@ using System.Linq;
 using System.Net;
 using System.Text;
 using System.Threading.Tasks;
-using Xunit;
 using Consul;
 using Newtonsoft.Json;
+using Xunit;
 
 namespace Consul.Test
 {
@@ -126,8 +126,8 @@ namespace Consul.Test
             {
                 Kind = "service-intentions",
                 Name = "Decepticon Assembler",
-                
-                
+
+
             };
 
 
--- a/Consul.Test/ConfigurationTest.cs
+++ b/Consul.Test/ConfigurationTest.cs
@@ -25,12 +25,19 @@ using System.Threading.Tasks;
 using Consul;
 using Newtonsoft.Json;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Consul.Test
 {
     [Collection(nameof(ExclusiveCollection))]
     public class ConfigurationTest : BaseFixture
     {
+        private readonly ITestOutputHelper _output;
+
+        public ConfigurationTest(ITestOutputHelper output)
+        {
+            _output = output;
+        }
         [Fact]
         public async Task Configuration_ApplyConfig()
         {
@@ -119,14 +126,71 @@ namespace Consul.Test
             var firstEntry = new ServiceIntentionsEntry
             {
                 Kind = "service-intentions",
-                Name = "Autobots Assembler",
+                Name = "Autobots-Assembler",
+                Sources = new List<SourceIntention>
+                {
+                    new SourceIntention
+                    {
+                        Name = "fortunate",
+                        Action = "allow"
+                    },
+                    new SourceIntention
+                    {
+                        Name = "Prad",
+                        Action = "allow"
+                    }
+                    //new SourceIntention
+                    //{
+                    //    Name = "!Optimus-Prime",
+                    //    Permissions = new List<IntentionPermission>
+                    //    {
+                    //       new IntentionPermission
+                    //       {
+                    //           Action = "allow",
+                    //           HTTP = new IntentionHTTPPermission
+                    //           {
+                    //               PathPrefix = "/",
+                    //               Methods = new List<string>{"GET", "POST"}
+                    //           }
+                    //       }
+                    //    }
+                    //}
+                }
             };
 
             var secondEntry = new ServiceIntentionsEntry
             {
                 Kind = "service-intentions",
-                Name = "Decepticon Assembler",
-
+                Name = "*",
+                Sources = new List<SourceIntention>
+                {
+                    new SourceIntention
+                    {
+                        Name = "middle",
+                        Action = "deny"
+                    },
+                    new SourceIntention
+                    {
+                        Name = "center",
+                        Action = "allow"
+                    }
+                    //new SourceIntention
+                    //{
+                    //    Name = "!Megatron",
+                    //    Permissions = new List<IntentionPermission>
+                    //    {
+                    //       new IntentionPermission
+                    //       {
+                    //           Action = "allow",
+                    //           HTTP = new IntentionHTTPPermission
+                    //           {
+                    //               PathPrefix = "/v2",
+                    //               Methods = new List<string>{"DELETE", "POST"}
+                    //           }
+                    //       }
+                    //    }
+                    //}
+                }
 
             };
 
@@ -137,13 +201,21 @@ namespace Consul.Test
             var resultTwo = await _client.Configuration.ApplyConfig(secondEntry);
             Assert.Equal(HttpStatusCode.OK, resultTwo.StatusCode);
 
-            var intentionsQuery = await _client.Configuration.ListIntentions();
+            var intentionsQuery = await _client.Configuration.ListIntentions<ServiceIntentionsEntry>();
             Assert.Equal(HttpStatusCode.OK, intentionsQuery.StatusCode);
 
             var intentions = intentionsQuery.Response;
             Assert.NotNull(intentions);
-            Assert.Contains(intentions, i => i.Name == firstEntry.Name);
-            Assert.Contains(intentions, i => i.Equals(secondEntry.Name));
+            //foreach (var intention in intentions)
+            //{
+            //    _output.WriteLine($"Intentions: {intention.Name}");
+            //    foreach (var source in intention.Sources)
+            //    {
+            //        _output.WriteLine($"Intentions: {source}");
+            //    }
+            //}
+            Assert.Contains(intentions, i => i.Name == "Autobots-Assembler");
+            Assert.Contains(intentions, i => i.Name == "*");
 
             await _client.Configuration.DeleteConfig(firstEntry.Kind, firstEntry.Name);
             await _client.Configuration.DeleteConfig(secondEntry.Kind, secondEntry.Name);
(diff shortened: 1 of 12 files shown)


Consul/Connect.cs
```diff
@@ -53,6 +54,178 @@ public class CAConfig
...
+        public bool Invert { get; set; }
+    }
+
+    /// <summary>
+    /// This handles the response for any operation carried out on the ServiceIntentionsEntry Model
+    /// </summary>
+    public class ServiceIntentionsEntryResponse
+    {
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public Guid ID { get; set; }
```
marcin-krystianc, line 185: We use `string` in other places, it is better to be consistent:
```suggestion
        public string ID { get; set; }
```

Consul/Connect.cs
```diff
@@ -107,6 +280,29 @@ public Task<WriteResult> CASetConfig(CAConfig config, WriteOptions q, Cancellati
...
+        /// </summary>
+        /// <param name="q">Custom query options</param>
+        /// <param name="ct">Cancellation Token</param>
+        /// <returns>A list of service intentions</returns>
+        ///
+        public Task<QueryResult<List<ServiceIntentionResponse>>> ListIntentions<ServiceIntentionResponse>(QueryOptions q, CancellationToken ct = default)
+        {
+            var req = _client.Get<List<ServiceIntentionResponse>>("/v1/connect/intentions", q);
+            return req.Execute(ct);
+
```
marcin-krystianc, line 304: nit: extra empty lien, please remove.

Consul.Test/ConnectTest.cs
```diff
@@ -76,5 +80,75 @@ public async Task Connect_CASetConfig()
...
+
+            var intentionsQuery = await _client.Connect.ListIntentions<ServiceIntentionsEntryResponse>();
+            Assert.Equal(HttpStatusCode.OK, intentionsQuery.StatusCode);
+
+            var intentions = intentionsQuery.Response;
+            Assert.NotNull(intentions);
+
+
+            Assert.Contains(intentions, i => i.SourceName == "fortunate");
+            Assert.Contains(intentions, i => i.DestinationName == firstEntry.Name);
```
marcin-krystianc, line 148: Please make the test more comprehensive by testing all the members of an intention.
If the exact expected value is unknown, we can at least test if all fields are non-null or non-zero.

marcin-krystianc (commented):
I've left some comments, please remove the erroneous definitions where suggested and add more assertions (as suggested) to the `Connect_ListIntention` test method.

Consul/Connect.cs
```diff
@@ -53,6 +54,178 @@ public class CAConfig
...
+        public string LegacyID { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public Dictionary<string, string> LegacyMeta { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public DateTime LegacyCreateTime { get; set; } = DateTime.UtcNow;
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public DateTime LegacyUpdateTime { get; set; } = DateTime.UtcNow;
```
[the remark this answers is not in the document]
FortunateOmonuwa: The time format sent when I tried to create an Intention was different from what was expected on the client.
So I felt since the time value was going to be assigned at the point of creation automatically anyways, I could just give it a default value so the format would match what the client expects

Consul.Test/ConnectTest.cs
```diff
@@ -76,5 +79,94 @@ public async Task Connect_CASetConfig()
...
+            Assert.Contains(intentions, i => !string.IsNullOrEmpty(i.SourceName));
+            Assert.Contains(intentions, i => !string.IsNullOrEmpty(i.DestinationNS));
+            Assert.Contains(intentions, i => !string.IsNullOrEmpty(i.SourceType));
+            Assert.Contains(intentions, i => !string.IsNullOrEmpty(i.SourceNS));
+            Assert.Contains(intentions, i => !string.IsNullOrEmpty(i.Action) && i.Action == "allow" || i.Action == "deny");
+            Assert.Contains(intentions, i => i.CreateIndex > 0);
+            Assert.Contains(intentions, i => i.ModifyIndex > 0);
+            Assert.Contains(intentions, i => i.Precedence > 0);
+            Assert.Contains(intentions, i => i.SourceName == "fortunate" && i.DestinationName == firstEntry.Name);
+            Assert.Contains(intentions, i => i.SourceName == "Optimus-Prime" && i.DestinationName == secondEntry.Name);
```
marcin-krystianc, line 166: Generally, these assertions are OK, but they are slightly more complicated than necessary.
It would be better to assign one of the intentions to a variable and then 
write assertions like this:
```
var int1 = intentions.First(i => i.SourceName == "Optimus-Prime");
Assert.NotEmpty(int1.DestinationName);
Assert.True(int1.CreateIndex > 0);
...
```

Consul.Test/ConnectTest.cs
```diff
@@ -76,5 +79,94 @@ public async Task Connect_CASetConfig()
...
+                    new SourceIntention
+                    {
+                        Name = "Prad",
+                        Action = "allow",
+                        LegacyCreateTime = DateTime.UtcNow,
+                        LegacyUpdateTime = DateTime.UtcNow,
+                    },
+                    new SourceIntention
+                    {
+                        Name = "Medhi",
```
marcin-krystianc, line 108:
```suggestion
                        Name = "Mehdi",
```

Consul.Test/ConfigurationTest.cs
```diff
@@ -22,7 +22,9 @@
 using System.Net;
 using System.Text;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
```
marcin-krystianc, line 25: Unnecessary, please remove.

Consul.Test/ConfigurationTest.cs
```diff
@@ -22,7 +22,9 @@
...
 using Xunit;
+using Xunit.Abstractions;
```
marcin-krystianc, line 27: Unnecessary, please remove.

Consul/Connect.cs
```diff
@@ -53,6 +54,173 @@ public class CAConfig
...
+        public string Partition { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public Dictionary<string, string> Meta { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public List<SourceIntention> Sources { get; set; }
+    }
+
+    public class SourceIntention
```
marcin-krystianc, line 83: I think it should be called `IntentionSource`, what do you think?

Consul/Connect.cs
```diff
@@ -53,6 +54,173 @@ public class CAConfig
...
+        public string Action { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public List<IntentionPermission> Permissions { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public int Precedence { get; set; }
+
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public string Type { get; set; } = "consul";
```
marcin-krystianc, line 107: Please don't set the default value here, it isn't needed.

Consul/Connect.cs
FortunateOmonuwa: it was named SourceIntention of the documentation 
https://developer.hashicorp.com/consul/docs/v1.11.x/connect/config-entries/service-intentions#sourceintention

Would you prefer i called it IntentionSource?

Consul/Configuration.cs
```diff
@@ -1773,6 +1747,7 @@ public Task<WriteResult> DeleteConfig(string kind, string name, CancellationToke
         {
             return DeleteConfig(kind, name, WriteOptions.Default, ct);
         }
+
```
marcin-krystianc, line 1750: Unnecessary change, please remove.

Consul/Connect.cs
```diff
@@ -53,6 +54,173 @@ public class CAConfig
         public ulong CreateIndex { get; set; }
         public ulong ModifyIndex { get; set; }
     }
+
+
```
marcin-krystianc, line 58: Unnecessary empty line, please remove one.

marcin-krystianc (commented):
There are only some formatting issues, looks good otherwise.

marcin-krystianc:
One more thing. Tests fail for older versions of consul, please use this technique to skip them when we are running tests for old versions -> [cutOffVersion](https://github.com/G-Research/consuldotnet/blob/master/Consul.Test/AgentTest.cs#L542-L546)

Consul.Test/ConnectTest.cs
```diff
@@ -76,5 +79,97 @@ public async Task Connect_CASetConfig()
             Assert.Equal("bar", updatedConfig.State["foo"]);
             Assert.Equal("", updatedConfig.Config["PrivateKey"]);
         }
+
+        [Fact]
+        public async Task Connect_ListIntentions()
+        {
+            var cutOffVersion = SemanticVersion.Parse("1.9.0");
+            Skip.If(AgentVersion < cutOffVersion, $"Current version is {AgentVersion}, but `logjson` is only supported from Consul {cutOffVersion}");
```
marcin-krystianc, line 87:
```suggestion
            Skip.If(AgentVersion < cutOffVersion, $"Current version is {AgentVersion}, but `service intentions` is only supported from Consul {cutOffVersion}");
```

Consul.Test/ConnectTest.cs
```diff
@@ -83,6 +83,9 @@ public async Task Connect_CASetConfig()
         [Fact]
```
marcin-krystianc, line 83:
```suggestion
        [SkippableFact]
```

marcin-krystianc (approved):
Nice work 👍

PR #433, opened 2025-03-02
Create intention with ID

FortunateOmonuwa:
This PR adds supports to [Create Intention With ID](https://developer.hashicorp.com/consul/api-docs/connect/intentions#create-intention-with-id)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
     strategy:
       matrix:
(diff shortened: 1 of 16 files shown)


marcin-krystianc:
Generally it looks ok, but I just noticed that "Create intention with ID" is deprecated (https://developer.hashicorp.com/consul/api-docs/connect/intentions#create-intention-with-id) and we should implement https://developer.hashicorp.com/consul/api-docs/connect/intentions#upsert-intention-by-name instead.

FortunateOmonuwa:
Yeah..
I intend to implement it too and create a PR today.

I implemented this first because it is the only endpoint returning the ID of a service intention and I need the ID for other implementations

PR #436, opened 2025-03-04
Read specific intention by name

FortunateOmonuwa:
This PR adds supports to [Read Specific Intention By Name](https://developer.hashicorp.com/consul/api-docs/v1.11.x/connect/intentions#read-specific-intention-by-name)

### DIFF
--- a/Consul.Test/ConnectTest.cs
+++ b/Consul.Test/ConnectTest.cs
@@ -226,5 +226,24 @@ namespace Consul.Test
             Assert.True(intention.Precedence > 0);
             await _client.Configuration.DeleteConfig("service-intentions", newEntry.DestinationName);
         }
+
+        [SkippableFact]
+        public async Task Connect_UpsertIntentionByName()
+        {
+            var cutOffVersion = SemanticVersion.Parse("1.9.0");
+            Skip.If(AgentVersion < cutOffVersion, $"Current version is {AgentVersion}, but setting CA config is only supported from Consul {cutOffVersion}");
+
+            var newEntry = new ServiceIntention
+            {
+                Action = "allow",
+                SourceType = "consul",
+            };
+
+            var req = await _client.Connect.UpsertIntentionsByName(newEntry, "LukaDoncic", "Lakers");
+            Assert.Equal(HttpStatusCode.OK, req.StatusCode);
+            Assert.True(req.Response);
+
+        }
+
     }
 }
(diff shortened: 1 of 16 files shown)


Consul.Test/ConnectTest.cs
```diff
@@ -173,6 +173,45 @@ public async Task Connect_ListIntentions()
...
+                        LegacyCreateTime = DateTime.UtcNow,
+                        LegacyUpdateTime = DateTime.UtcNow,
+                    }
+                }
+            };
+
+            var req = await _client.Configuration.ApplyConfig(newEntry);
+            Assert.Equal(HttpStatusCode.OK, req.StatusCode);
+
+            var sourceName = newEntry.Sources.First(x => x.Name == "Kanye West").Name;
```
marcin-krystianc, line 200: It looks a little bit clunky. 
You could define `sourceName` and 'destinationName' at the beginning of  the test - it is going to be simpler.

marcin-krystianc (commented):
Looks good, just small remark about test readability and complexity.

Consul.Test/ConnectTest.cs
FortunateOmonuwa: Alright...
But i think i might need to close this PR and reopen another one for this implementation

I noticed the test my upsertintentionbyname implementation which has been merged is missing a [SkippableFact] attribute and I don't want to make any changes to it.

So I want to create a new branch for the implementation so I can have the right thing there.

Or do you have a better way I can do this please???

marcin-krystianc: > Alright... But i think i might need to close this PR and reopen another one for this implementation
> 
> I noticed the test my upsertintentionbyname implementation which has been merged is missing a [SkippableFact] attribute and I don't want to make any changes to it.
> 
> So I want to create a new branch for the implementation so I can have the right thing there.
> 
> Or do you have a better way I can do this please???

Good catch with the missing `[SkippableFact]` for `Connect_UpsertIntentionByName`. You can keep this PR open, and just create another branch (branch of the current main branch) and add the missing `[SkippableFact]` attribute and open a PR.
It is ok, to have multiple PRs changing code that is nearby,  in worst case that will create some merge conflicts that will need to be resolved on your branch..

FortunateOmonuwa: The thing is, on the Master branch I just checked again to confirm, it has the [SkippableFact] attribute.
So I think I made the mistake while trying to resolve the conflicts.

Maybe I could just reset back to the point before resolving the conflict. 
Is this going to work?

PR #442, opened 2025-03-06
Read Specific Intention By Name

FortunateOmonuwa:
This PR adds supports to [Read Specific Intention By Name](https://developer.hashicorp.com/consul/api-docs/v1.11.x/connect/intentions#read-specific-intention-by-name)

### DIFF
--- a/Consul.Test/ConnectTest.cs
+++ b/Consul.Test/ConnectTest.cs
@@ -244,5 +244,50 @@ namespace Consul.Test
             var deletedIntentionExists = allIntentions.Response.Any(x => x.SourceName == newEntry.SourceName);
             Assert.False(deletedIntentionExists);
         }
+
+        [SkippableFact]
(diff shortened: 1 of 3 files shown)


Consul.Test/ConnectTest.cs
```diff
@@ -208,5 +208,50 @@ public async Task Connect_UpsertIntentionByName()
...
+                    {
+                        Name = "Kanye West",
+                        Action = "deny",
+                        LegacyCreateTime = DateTime.UtcNow,
+                        LegacyUpdateTime = DateTime.UtcNow,
+                    }
+                }
+            };
+
+            var source = newEntry.Sources.First(x => x.Name == "Kanye West").Name;
```
marcin-krystianc, line 234: Please define the source name and destination name as variables, that will let you get rid of this `Sources.First(x => x.Name == "Kanye West").Name` which is clunky.

Issue #353, opened 2024-07-18
Fix some flaky tests:


marcin-krystianc:
### Describe the bug

[ ] - Health_GetLocalNode
[ ] - Semaphore_OneShot (https://github.com/G-Research/consuldotnet/actions/runs/9988424558/job/27604909349)
[ ] - Health_Connect (https://github.com/G-Research/consuldotnet/actions/runs/9987899060/job/27603280179)
[ ] - Catalog_GetNodesForMeshCapableService

### Steps To Reproduce

These tests fail occasionaly

### Expected behavior

Tests shouldn't randomly fail

### Environment

No specific environment

PR #527, opened 2025-11-06
bug fix: attempt to fix flaky test

Davidcode-png:
## Description
This PR introduces a fix for flaky test on the CI pipeline caused by a race condition when querying nodes

## Related Issues

Addresses #353 
## Checklist

Please make sure to check the following before submitting your pull request:

- [x] Did you read the [contributing guidelines](https://consuldot.net/docs/contributing/guidelines)?
- [ ] Did you update the docs?
- [x] Did you write any tests to validate this change?

### DIFF
--- a/Consul.Test/CatalogTest.cs
+++ b/Consul.Test/CatalogTest.cs
@@ -230,20 +230,16 @@ namespace Consul.Test
             };
 
             await _client.Agent.ServiceRegister(registration);
-            // First request - no WaitIndex set
+            // First request - with no WaitIndex set
             var services = await _client.Catalog.NodesForMeshCapableService(registration.Name);
             if (services.Response == null || services.Response.Length == 0)
             {
-                //Assert.NotEmpty(services.Response);
-                //Assert.Equal(services.Response[0].ServiceID, registration.Name + "-sidecar-proxy");
                 // Get the index from the first response
-                var currentIndex = services.LastIndex;
-                var options = new QueryOptions() { WaitIndex = currentIndex, };
+                var options = new QueryOptions() { WaitIndex = services.LastIndex, };
                 services = await _client.Catalog.NodesForMeshCapableService(registration.Name, options);
             }
             Assert.NotEmpty(services.Response);
             Assert.Equal(services.Response[0].ServiceID, registration.Name + "-sidecar-proxy");
-
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


Consul.Test/CatalogTest.cs
```diff
@@ -235,7 +235,14 @@ public async Task Catalog_GetNodesForMeshCapableService()
             };
             await _client.Agent.ServiceRegister(registration);
 
-            var services = await _client.Catalog.NodesForMeshCapableService(registration.Name);
+            var options = new QueryOptions()
+            {
+                WaitIndex = 1000,
+                WaitTime = new TimeSpan(0, 0, 5)
+            };
+            await _client.Agent.ServiceRegister(registration);
```
marcin-krystianc: The `ServiceRegister` is being called twice; it should be called once.

Davidcode-png: Will update that.

Consul.Test/CatalogTest.cs
```diff
@@ -233,11 +233,22 @@ public async Task Catalog_GetNodesForMeshCapableService()
...
+            // First request - no WaitIndex set
             var services = await _client.Catalog.NodesForMeshCapableService(registration.Name);
             Assert.NotEmpty(services.Response);
             Assert.Equal(services.Response[0].ServiceID, registration.Name + "-sidecar-proxy");
+            // Get the index from the first response
+            var currentIndex = services.LastIndex;
+            var options = new QueryOptions()
+            {
+                WaitIndex = currentIndex,
+                WaitTime = new TimeSpan(0, 0, 5)
```
marcin-krystianc: Wait time is not needed, we just want to wait for an update.

marcin-krystianc (commented):
It is going in the right direction, but it would solve the problem if it is done like this. Before sending the second request we need to see if the first response is satisfactory or not (with if condition instead of assertion). We should send the second request only if the first reposes didn't have the information we are expecting.

Consul.Test/CatalogTest.cs
```diff
@@ -225,19 +225,25 @@ public async Task Catalog_GetNodesForMeshCapableService()
...
             };
-            await _client.Agent.ServiceRegister(registration);
 
+            await _client.Agent.ServiceRegister(registration);
+            // First request - no WaitIndex set
             var services = await _client.Catalog.NodesForMeshCapableService(registration.Name);
+            if (services.Response == null || services.Response.Length == 0)
+            {
+                //Assert.NotEmpty(services.Response);
+                //Assert.Equal(services.Response[0].ServiceID, registration.Name + "-sidecar-proxy");
```
marcin-krystianc:
```suggestion
```

Consul.Test/CatalogTest.cs
```diff
@@ -225,19 +225,25 @@ public async Task Catalog_GetNodesForMeshCapableService()
...
+                // Get the index from the first response
+                var currentIndex = services.LastIndex;
```
marcin-krystianc: No need to instantiate  a new 'currentIndex' variable here. It doesn't add any value.

PR #544, opened 2026-01-10
Get rid of Lazy<>

sungam3r:
## Description

`Lazy<>` fields essentially give nothing. All "resources" inside their instances are just tiny objects, no long init operations at all. Regarding saving memory - memory is already wasted with Lazy instances. It's just cheaper to use all wrapped objects directly.

## Checklist

Please make sure to check the following before submitting your pull request:

- [ ] Did you read the [contributing guidelines](https://consuldot.net/docs/contributing/guidelines)?
- [ ] Did you update the docs?
- [ ] Did you write any tests to validate this change?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+## 1.7.14.10
+* Add create ACL Binding Rules (#506)
+* Add raft transfer leader (#489)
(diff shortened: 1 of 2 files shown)


sungam3r:
Lazy was introduced just as shorter replacement of already existed construction - lock with double check. The real question is what this construction accomplished before.

PR #640, opened 2026-06-12
Add zizmor workflow for GitHub Actions static analysis

david-wagih:
## What

Adds a workflow that runs [zizmor](https://docs.zizmor.sh) — a static analyzer for GitHub Actions workflows — via the upstream [`zizmorcore/zizmor-action`](https://github.com/zizmorcore/zizmor-action) (SHA-pinned, v0.5.6), uploading results as SARIF to Code Scanning.

## Why

zizmor catches common CI security issues (excessive permissions, credential persistence, template injection, unpinned/impostor actions, cache poisoning). This complements the existing CodeQL setup, which covers the C# code but not the workflows themselves.

A current offline run against `master` (`zizmor 1.25.2, --min-severity low --min-confidence low`) reports **26 findings (18 medium, 8 high)** across the existing workflows — these would become visible in the Security tab for triage after merge.

## Design choices

- **Non-blocking**: the action does not fail the build on findings; they surface in the Security tab and as PR annotations.
- **Trigger shape mirrors the CodeQL workflow** (`push` to `master` + `pull_request`), paths-filtered to `.github/**` since that's all zizmor analyzes; `workflow_dispatch` for manual runs.
- **Least privilege**: top-level `permissions: {}`, job grants only `actions: read`, `contents: read`, `security-events: write`.
- **SHA-pinned** actions with version comments, consistent with the repo's Dependabot-managed pinning.
- The new workflow itself audits clean (zero zizmor findings).

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -7,13 +7,8 @@ name: zizmor
 on:
   push:
     branches: [master]
-    paths:
-      - ".github/**"
   pull_request:
     branches: [master]
-    paths:
-      - ".github/**"
-  workflow_dispatch:
 
 permissions: {}
 
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,0 +1,37 @@
+name: zizmor
+
+# Static analysis of GitHub Actions workflows with zizmor
+# (https://docs.zizmor.sh). Results are uploaded as SARIF to
+# Code Scanning; findings do not fail the build.
+
+on:
+  push:
+    branches: [master]
+    paths:
+      - ".github/**"
+  pull_request:
+    branches: [master]
+    paths:
+      - ".github/**"
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  zizmor:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write # SARIF upload to Code Scanning
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
+        with:
+          min-severity: low
+          min-confidence: low


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  push:
+    branches: [master]
+    paths:
+      - ".github/**"
+  pull_request:
+    branches: [master]
+    paths:
+      - ".github/**"
+  workflow_dispatch:
```
dgl: I think later manual runs could be useful (particularly for online checks), but for now it's just attack surface. Let's just remove that and think about it later.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,37 @@
...
+      security-events: write # SARIF upload to Code Scanning
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
+        with:
```
dgl, line 30: Given this triggers on `path: ".github/**"` I think we should set `inputs: ".github" otherwise the results could be inconsistent. Or maybe we should drop the path filter (because always running zizmor is potentially useful for online checks).

(The upstream docs don't suggest a path filter at all.)

dgl:
The code scanning results are working on this PR: https://github.com/G-Research/consuldotnet/pull/640/checks?check_run_id=81359056177

.github/workflows/zizmor.yml
marcin-krystianc: > Given this triggers on `path: ".github/**"` I think we should set `inputs: ".github" otherwise the results could be inconsistent. Or maybe we should drop the path filter (because always running zizmor is potentially useful for online checks).
> 
> (The upstream docs don't suggest a path filter at all.)

If it is not too "expensive" to run the workflow, let's remove the path filter for simplicity.

.github/workflows/zizmor.yml
david-wagih: Removed `workflow_dispatch` — agreed it's just attack surface for now, can revisit later if manual/online runs become useful.

.github/workflows/zizmor.yml
david-wagih: Dropped the `paths: .github/**` filter on both push and pull_request, per the discussion. zizmor now runs on every push/PR to master (matching upstream guidance, and useful for online checks), with the default whole-repo input so results stay consistent.

dgl (approved):

